PR #9, opened 2021-04-19
Describe -msave-restore runtime library support

edward-jones:
As far as I am aware the implementation of `-msave-restore` has not been documented anywhere. I'm unsure whether this or https://github.com/riscv/riscv-c-api-doc is the more appropriate place to document this.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -207,6 +207,28 @@ RV32E. There is not currently a way to specify an alternative stack alignment,
 but the `-mpreferred-stack-boundary` and `-mincoming-stack-boundary` flags 
 supported by GCC on X86 could be adopted.
 
+## Save restore support
+
+The save restore optimization is enabled through the option `-msave-restore`
+and reduces the amount of code in the prologue and epilogue by using
+library functions instead of inline code to save and restore callee saved
+registers. The library functions are provided in the emulation library and
+have the following signatures:
+
+* `void __riscv_save_<N>(void)`
+* `void __riscv_restore_<N>(void)`
+
+`<N>` is a value between 0 and 12 and corresponds to the number of
+registers between `s0` and `s11` that are saved/restored. The return
+address register `ra` is always included in the registers saved and restored.
+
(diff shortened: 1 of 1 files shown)


README.mkd
```diff
@@ -207,6 +207,27 @@ RV32E. There is not currently a way to specify an alternative stack alignment,
...
+* `void __riscv_restore_<N>(void)`
+
+`<N>` is a value between 0 and 12 and corresponds to the number of
+registers between `s0` and `s11` that are saved/restored. The return
+address register `ra` is always included in the registers saved and restored.
+
+The `__riscv_save_<N>` functions are called from the prologue. They allocate
+stack space for the registers and then save `ra` and the appropriate number
+of registers from `s0`-`s11`. The `__riscv_restore_<N>` functions are called
+from just before the return instruction in the epilogue and restore the
```
jim-wilson, line 228: The restore functions are tail-called instead of the return instruction.  The function return happens inside the restore function.

README.mkd
```diff
@@ -207,6 +207,27 @@ RV32E. There is not currently a way to specify an alternative stack alignment,
...
+have the following signatures:
+
+* `void __riscv_save_<N>(void)`
+* `void __riscv_restore_<N>(void)`
+
+`<N>` is a value between 0 and 12 and corresponds to the number of
+registers between `s0` and `s11` that are saved/restored. The return
+address register `ra` is always included in the registers saved and restored.
+
+The `__riscv_save_<N>` functions are called from the prologue. They allocate
```
jim-wilson, line 225: The save function call must use the alternate link register t0 instead of the regular ra, to avoid clobbering ra.  The save function then returns to the address in t0.

jim-wilson (commented):
I don't care whether this goes in toolchain-conventions or c-abi.  We probably should merge those two documents.

edward-jones:
I've updated to clarify that the save-function uses `t0` and that the restore functions are tail-called. Hopefully this is sufficiently clear now.

jim-wilson (approved):

kito-cheng:
@asb

PR #14, opened 2021-06-22
Relax the ISA string order checking for -march

kito-cheng:
Changes:
2021/08/16: Remove order constraint between single-letter and multi-letter, also add 2 more NOTE.

---

It's corresponding for https://github.com/riscv/riscv-toolchain-conventions/issues/11 ,

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -57,12 +57,49 @@ e.g. `-march=rv64g`. A target `-march` which includes floating point
 instructions implies a hardfloat calling convention, but can be overridden 
 using the `-mabi` flag (see the next section).
 
-The ISA subset naming conventions are described in Chapter 27 of the RISC-V 
-user-level ISA specification. However, tools do not currently follow this 
-specification (input is case sensitive, ...).
+The ISA subset naming conventions and canonical order are described in
+Chapter `ISA Extension Naming Conventions` of the RISC-V user-level ISA
+specification. However, tools do not currently follow this specification
+(input is case sensitive, ...).
+
+The rule of ISA string become more complicated, due to extensions implication
+rule and more extensions added into RISC-V, the canonical order is non-obvious
+to human, so tools can accept the ISA string in non-canonical order for reduce
+the burden of remembering the canonical order.
+
+Detail rule for ISA string:
+
+1. First letter must be `i`, `e` or `g`.
+2. Single-letter may be non-canonical order.
+3. Multi-letter may be non-canonical order.
+4. Multi-letter must be separated by underscore.
+5. Version separator(`p`) has higher priority than `p` extension.
+
+Example:
+```
+rv32ima_zicsr        # Valid ISA string.
+rv32i_zicsr_m        # Valid ISA string.
+rv32i_zicsr_ma       # Valid ISA string.
+rv32imac             # Valid ISA string.
+rv32mai              # Invalid ISA string, first letter must be `i`, `e` or `g`.
+rv32i_zicsrzifence   # Valid ISA string, but it will interpreted as rv32
+                     # with base extension and `zicsrzifence` extension
+                     # rather than `zicsr` and `zifence` extensions.
+rv32i2p1             # Valid ISA string, it recognized as `I` extension with
+                     # version 2.1 rather than `I` extension with with version
+                     # 2.0 and `P` extension with 1.0.
+```
 
 If the 'C' (compressed) instruction set extension is targeted, the compiler 
 will generate compressed instructions where possible.
+
+NOTE: Single-letter extension with version (e.g. `m2p0`) still treat as
+      single-letter extension, we won't treat it as multi-letter extension.
+
+NOTE: Any output of ISA string like `Tag_RISCV_arch` must be canonical order.
+
+NOTE: Cross-tool argument are highly recommended passed in canonical order for
+      backward compatible.
 
 ### Issues for consideration
 * Whether `riscv32` and `riscv64` should be accepted as synonyms for `rv32` 


README.mkd
```diff
@@ -54,10 +54,27 @@ base and extensions, e.g. `-march=rv64g`. A target `-march` which includes
...
+specification. However, tools can accept the ISA string in non-canonical
+order for reduce the burden of remembering the canonical order.
+
+Detail rule for ISA string:
+
+1. First letter must be `i`, `e` or `g`.
+2. Multi-letter extension must come after single-letter extension.
+3. Single-letter could be non-canonical order.
+4. Multi-letter could be non-canonical order.
+5. Multi-letter must seperated by underscore.
```
jim-wilson, line 68: seperated -> separated

jim-wilson (approved):
with a minor typo fixed

README.mkd
kito-cheng: Thanks :)

README.mkd
```diff
@@ -54,10 +54,27 @@ base and extensions, e.g. `-march=rv64g`. A target `-march` which includes
...
+The ISA subset naming conventions and canonical order are described in
+Chapter `ISA Extension Naming Conventions` of the RISC-V user-level ISA
+specification. However, tools can accept the ISA string in non-canonical
+order for reduce the burden of remembering the canonical order.
+
+Detail rule for ISA string:
+
+1. First letter must be `i`, `e` or `g`.
+2. Multi-letter extension must come after single-letter extension.
+3. Single-letter could be non-canonical order.
```
ebahapo, line 66: `s/could/may/`

README.mkd
```diff
@@ -54,10 +54,27 @@ base and extensions, e.g. `-march=rv64g`. A target `-march` which includes
...
+4. Multi-letter could be non-canonical order.
```
ebahapo, line 67: Ditto.

README.mkd
```diff
@@ -54,10 +54,27 @@ base and extensions, e.g. `-march=rv64g`. A target `-march` which includes
...
+5. Multi-letter must separated by underscore.
```
ebahapo, line 68: "must **be** separated"

README.mkd
```diff
@@ -54,10 +54,27 @@ base and extensions, e.g. `-march=rv64g`. A target `-march` which includes
...
+
+Example:
+```
+rv32ima_zicsr        # Valid ISA string.
+rv32mai              # Invalid ISA string, first letter must be `i`, `e` or `g`.
+rv32i_zicsr_m        # Invalid ISA string, multi-letter extension must come after single-letter extension.
+rv32i_zicsrzifence   # Valid ISA string, but it will interpreted as rv32 with base extension and
```
ebahapo, line 75: Also, I think that this is unreadable.  Multi letter extensions should be separated by underscores from each other too.

ebahapo:
It should also be specified that the tools always emit the extensions in canonical order.

kito-cheng:
Changes:
- Remove order constraint between single-letter 
- Add 2 NOTE
  - ELF attribute must be canonical order
  - Cross tool argument strongly recommend passed in canonical order, e.g. GCC invoke gnu assembler and LLVM with gnu assembler.

README.mkd
```diff
@@ -54,14 +54,40 @@ base and extensions, e.g. `-march=rv64g`. A target `-march` which includes
...
+The ISA subset naming conventions and canonical order are described in
+Chapter `ISA Extension Naming Conventions` of the RISC-V user-level ISA
+specification. However, tools can accept the ISA string in non-canonical
+order for reduce the burden of remembering the canonical order.
+
+Detail rule for ISA string:
+
+1. First letter must be `i`, `e` or `g`.
+2. Multi-letter extension must come after single-letter extension.
+3. Single-letter may be non-canonical order.
```
rjiejie, line 66: we need to consider the single-letter with version like as m1p1,
should it be treated as a single character wholly ?

kito-cheng: Yeah, I would like to consider single-letter with version as normal single-letter ext. because following reason:
1. ISA manual has example usage like: `RV64I1p0M1p0A1p0F1p0D1p0` and `RV64I1M1A1F1D1`, so I prefer to make those still valid input for toolchain
2. Current arch string parser can handle that, although might be ambiguous when P extension is present.

Anyway, I think we could add note about that :)

README.mkd
```diff
@@ -54,14 +54,40 @@ base and extensions, e.g. `-march=rv64g`. A target `-march` which includes
...
+4. Multi-letter may be non-canonical order.
+5. Multi-letter must be separated by underscore.
+
+Example:
+```
+rv32ima_zicsr        # Valid ISA string.
+rv32i_zicsr_m        # Valid ISA string.
```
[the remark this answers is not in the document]
kito-cheng: I forgot to remove that at last update, I intend to remove rule 2 but seems like I only update the example...

README.mkd
```diff
@@ -54,14 +54,40 @@ base and extensions, e.g. `-march=rv64g`. A target `-march` which includes
...
+
+1. First letter must be `i`, `e` or `g`.
+2. Multi-letter extension must come after single-letter extension.
+3. Single-letter may be non-canonical order.
+4. Multi-letter may be non-canonical order.
+5. Multi-letter must be separated by underscore.
+
+Example:
+```
+rv32ima_zicsr        # Valid ISA string.
```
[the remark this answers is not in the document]
kito-cheng: Add one more rule for disambiguate version seperator and 'p' ext., for zp*... I would like to wait ISA spec has some clarification (e.g. https://github.com/riscv/riscv-isa-manual/pull/718), before that, let keep vague here.

README.mkd
```diff
@@ -54,14 +54,40 @@ base and extensions, e.g. `-march=rv64g`. A target `-march` which includes
 floating point instructions implies a hardfloat calling convention, but can be 
 overridden using the `-mabi` flag (see the next section).
 
-The ISA subset naming conventions are described in Chapter 22 of the RISC-V 
-user-level ISA specification. However, tools do not currently follow this
```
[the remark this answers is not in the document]
kito-cheng: Added back.

README.mkd
rjiejie: Sounds good :)

kito-cheng:
Changes:
- Remove rule 2 
- Add one more rule for disambiguate version separator and 'p' ext.
- Add note for single letter ext. with version

lazyparser:
Hi @kito-cheng 

Would you mind to update the status or progress when available?

MaskRay (approved):
It looks like there is a GCC patch dropping the order requirement
https://gcc.gnu.org/pipermail/gcc-patches/2024-January/642151.html

If we had fewer extensions, ensuring a canonical order is better as a
code search of a fixed string will retrieve the relevant results, and
I'd wish that we did not lose the strictness.
Now that there are a hundred extensions, and there are very bizarre use cases, I agree that enforcing a strict order has lost its goodness...

README.mkd
```diff
@@ -54,14 +54,53 @@ base and extensions, e.g. `-march=rv64g`. A target `-march` which includes
...
+4. Multi-letter must be separated by underscore.
+5. Version separator(`p`) has higher priority than `p` extension.
+
+Example:
+```
+rv32ima_zicsr        # Valid ISA string.
+rv32i_zicsr_m        # Valid ISA string.
+rv32imac             # Valid ISA string.
+rv32mai              # Invalid ISA string, first letter must be `i`, `e` or `g`.
+rv32i_zicsrzifence   # Invalid ISA string, because it will interpreted as rv32
```
asb, line 81: Is "invalid" the right description here? Isn't it a valid string, but the user probably didn't mean to specify an extension named "zicsrzifence"

kito-cheng: Yeah, that should be also valid but will interpreted as zicsrzifence, updated :)

kito-cheng:
GCC patch has bee merged into trunk :)

cmuellner:
I think this is ready to land.

kito-cheng:
Ohhh, we made it before 2024 lunar new year 🎉

PR #15, opened 2021-07-22
Describe the behaviour of the linker for ABI mismatches

compnerd:
Certain ABI requirements of the code contained in an object file are
described by the `e_flags` field of the ELF header in an ELF object
file.  An ELF linker is able to inspect this and provide validation that
ABI mismatches do not accidentally occur.

Describe how the mismatches should be validated so that different
toolchains perform the validation similarly.  Currently the BFD linker
is more lenient about the check, permitting inconsequential object files
(files which do not contribute to the text segment) to be excluded from
the requirements while lld is overly restrictive and does not permit the
mismatch for any input.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -156,6 +156,31 @@ avoid this problem.
 pseudoinstructions and whether compressed instructions are printed directly or 
 not?
 
+## Linker behaviour
+
+The RISC-V architecture supports multiple ABI conventions. Some of them are
+incompatible with one another. Two particular cases of such irreconcilable
+differencces are the floating point passing convention and whether ILP32E
+calling convention (RVE) is in use. In order to identify (ELF) object files
+built with these ABI constraints, the `e_flags` field in the ELF header contains
+a flag set to indicate the ABI in use. The details of the flag can be found in
+the [RISCV
+psABI](https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md#file-header).
+
+The linker can detect a mismatch of the ABI and indicate to the developer of the
+conflict. A mismatch should be flagged as an irrecoverable error. However,
+because the impact of the calling convention is limited to the code
+contributions of an object file, any empty object file (i.e. containing no
+sections) or any object file which does not contribute any text sections which
+are loaded at runtime shall be excluded from the requirement that the floating
+point passing convention and the use of RVE match across all the object files.
+
+### Issues for consideration
+* If there is a usecase for multiple floating point passing conventions to be
+used in a single compiled image, it may be useful to consider lifting the
+restriction that the floating point calling convention shall match across all
+object files.
+
 ## Assembler behaviour
 
 See the [RISC-V Assembly Programmer's 


compnerd:
CC: @asb @jim-wilson

asb:
@jrtc27 @kito-cheng: any thoughts on this one?

jrtc27:
I dispute the "overly restrictive" part; you can come up with C inputs that aren't compatible yet BFD won't give an error for because it's trying to support other use cases that can be done an equivalent way (turning a binary blob into an ELF; historically people do things like `ld -b binary` or `objcopy -I binary -O elfNN-foo`, with no way to pass -march/-mabi information that then gets recorded in the output, but these days you can use `.incbin` in an assembly file as a much more flexible and robust version).

asb:
+@maskray too

jim-wilson (approved):

README.mkd
```diff
@@ -156,6 +156,31 @@ avoid this problem.
...
+a flag set to indicate the ABI in use. The details of the flag can be found in
+the [RISCV
+psABI](https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md#file-header).
+
+The linker can detect a mismatch of the ABI and indicate to the developer of the
+conflict. A mismatch should be flagged as an irrecoverable error. However,
+because the impact of the calling convention is limited to the code
+contributions of an object file, any empty object file (i.e. containing no
+sections) or any object file which does not contribute any text sections which
+are loaded at runtime shall be excluded from the requirement that the floating
```
jrtc27, line 175: To be clear: this is not what LLD does and I am personally of the view it should continue to not do this. Just because BFD does something doesn't mean it's the right thing to do. There are ways to avoid this issue entirely and I am loathe to provide users with more footguns than necessary in the LLVM ecosystem, regardless of what GNU developers chose to do 5 years ago.

README.mkd
```diff
@@ -156,6 +156,31 @@ avoid this problem.
...
+point passing convention and the use of RVE match across all the object files.
+
+### Issues for consideration
```
jrtc27, line 178: I think this should go away entirely. It's certainly not normative text, and linking files with different floating point ABIs that have code in them is definitely asking for so much trouble that you can't just disable the checking. This is how you end up with users building programs that blow up and complaining at toolchain developers that the compiler doesn't work.

cmuellner:
This was discussed in today's SIG Toolchain call.

The `e_flags` are documented in the [psABI repo](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc). There, we have documented that linkers disallow ELF files with incompatible ABIs:
> These flags are used by the linker to disallow linking ELF files with incompatible ABIs together

There is no formal definition of "incompatible ABIs". Defining ABI compatibility would result in a significant amount of work (with numerous corner cases and exceptions) with little benefit over keeping this as linker-specific behavior.

Besides that, it is unlikely that there will be broad agreement on exact ABI compatibility rules (especially since BFD and LLD already implement different compatibility rules). And a set of ABI compatibility rule recommendations would have little value (readers would not be able to rely on that behavior).

Therefore, I'm closing this.

PR #17, opened 2022-01-10
Conventions for vendor extension

kito-cheng:
Many RISC-V vendor has their own extensions, and some vendor are trying
to upstream their extension, so I think it would be great we can have a
common rule to make LLVM and GNU toolchain community are on the same page
here.


kito-cheng:
@asb @jim-wilson @Nelson1225 @cmuellner @aswaterman @palmer-dabbelt

README.mkd
```diff
@@ -229,6 +229,41 @@ registers and then save `ra` and the appropriate number of registers from
 epilogue. They restore the saved registers, deallocate the stack space for the
 register, and then perform a return through the restored value of `ra`.
 
+## Conventions for vendor extension
+
+Custom instruction is an important part of RISC-V, offical ISA spec has reserved
```
asb, line 234: "Support for custom instruction set extensions are an important part of RISC-V, with large encoding spaces reserved of vendor extensions."

README.mkd
```diff
@@ -229,6 +229,41 @@ registers and then save `ra` and the appropriate number of registers from
...
+serveral encoding space for extending instructions.
+
+However there is no offical rule on how to use those encoding space and the rule
```
asb, line 237: "However, there are no official guidelines on how to use this encoding space or on naming the mnemonics. This section defines guidelines which vendors are expected to follow if upstreaming support for their extensions. Although vendor-provided toolchains are free to make different choices, they are strongly urged to align with these guidelines in order to ensure there is a straightforward path for upstreaming in the future."

README.mkd
```diff
@@ -229,6 +229,41 @@ registers and then save `ra` and the appropriate number of registers from
...
+of assembly mnemonic, so this section is defining those rule to make it easier
+to management on upstream open source toolchain.
+
+### Vendor extension naming scheme
+
+According RISC-V ISA spec, non-standard extensions are named using a single `x`
+followed by an alphabetical name and an optional version number.
+
+For easier to identify and prevent naming conflict, vendor extension naming must
```
asb, line 246: "For easier to" => "To make it easier to"

asb (commented):
Thanks for kicking this off Kito. I've left a number of comments, which admittedly is a mix of grammatical changes / wording tweaks and what I think are policy clarifications (but it's possible in some cases I'm differing from your intent). So take a look and see what you think, and pick which bits you're happy with.

@ptomisch's binutils patch lists having publicly available documentation as a requirement. It's probably worth adding "Patches to add support for open source toolchains are unlikely to be accepted without specification documents being made available publicly." I realise I'm hedging my words a bit on that one - it's very hard to imagine patches being accepted to LLVM without such documentation, but I don't feel it's right for me to define policy on behalf of the whole LLVM project.

README.mkd
```diff
@@ -229,6 +229,41 @@ registers and then save `ra` and the appropriate number of registers from
...
 register, and then perform a return through the restored value of `ra`.
 
+## Conventions for vendor extension
+
+Custom instruction is an important part of RISC-V, offical ISA spec has reserved
+serveral encoding space for extending instructions.
+
+However there is no offical rule on how to use those encoding space and the rule
+of assembly mnemonic, so this section is defining those rule to make it easier
+to management on upstream open source toolchain.
```
ptomsich, line 239: "to management" -> "to manage"

README.mkd
ptomsich: I don't know if we want to talk about the 'encoding space' here, as we are not establishing any guidelines on how to use/subdivide/etc. the encoding space.

README.mkd
```diff
@@ -229,6 +229,41 @@ registers and then save `ra` and the appropriate number of registers from
...
+For example:
+- `XVentanaCondOps` from Ventana
+- `Xsfcflushdlone` from SiFive
+- `Xtheade` from T-head
+
+### Assembly mnemonic
+
+In order to avoid confusion between standard extension and other vendor
+extensions, vendor extension must prefixed with their own vendor prefix,
+e.g. `sf.` for SiFive, `vt.` for Ventana. and no registration required
```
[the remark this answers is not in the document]
ptomsich: Let's require that vendors register their extensions by listing them in this document.
If something outrageous or unadvisable — e.g. because it has an offensive meaning in some geography (the fact that every class defined in NextStep starts with 'NS' still makes me cringe) or language — comes along, this will give the community a chance to raise their concerns).

Let's add a table for this in the document:

### List of vendor prefixes

Prefix  | Vendor | URL
:------ | :------ | :------
vt | Ventana Micro Systems | https://www.ventanamicro.com/
sf | SiFive | https://www.sifive.com/

ptomsich (changes requested):

frasercrmck:
This is a good idea, thanks.

README.mkd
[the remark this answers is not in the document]
joshuascheid: I suggest a primary prefix so that there is no concern with vendor prefixes polluting or reducing the standard mnemonic name space.  Such could be `x` to match the extension prefix, or something like `vnd` for "vendor".

- `x.vt.hcf`
- `xvt.hcf`
- `vnd.vt.hcf`

kito-cheng:
Changes:
- Apply review comment from @asb 
- Add list of vendor prefixes

README.mkd
```diff
@@ -229,6 +229,57 @@ registers and then save `ra` and the appropriate number of registers from
...
+- `Xtheade` from T-head
+
+### Assembly mnemonic
+
+In order to avoid confusion between standard extension and other vendor
+extensions, instruction mnemonics from vendor extensions must have a
+prefix corresponding to the vendor's name.
+
+e.g. `sf.` for SiFive, `vt.` for Ventana. No central registration with RISC-V
+International or elsewhere is required before the prefix is used.
```
luismarques, line 263: Please add a comment that, although no centralized registration is required, they should add the vendor prefix to the table below, with a link to the section.

rjiejie:
@kito-cheng Looks good for me, it's helpful also for community & vendor,
we need to make a patch for early compatibility of user's binary codes :P

kito-cheng:
@rjiejie thanks! hope we can making step forward this year for vendor extension support on upstream!

README.mkd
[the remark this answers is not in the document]
rjiejie: > I suggest a primary prefix so that there is no concern with vendor prefixes polluting or reducing the standard mnemonic name space. Such could be `x` to match the extension prefix, or something like `vnd` for "vendor".
> 
> * `x.vt.hcf`
> * `xvt.hcf`
> * `vnd.vt.hcf`

@kito-cheng +1 for x.<vendor insn>, I think some user's interface more import like -march=*_xtheadc than internal name, we should make simple rule with assemble's name with existing rule "X"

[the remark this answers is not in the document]
kito-cheng: I am OK with `x.<vendor>.` prefix, @cmuellner @ptomsich @asb, any comment?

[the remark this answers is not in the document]
ptomsich: I find this a bit unwieldy and would prefer to use the vendor-prefix only (i.e. vt/th/…) instead of further prefixing (one of the reasons being that I like my disassembly to be nicely aligned, which gets harder if some mnemonics become very long).  This may also get quite uncomfortable, if vendors then use the appropriate '.uw' postfix on some instruction (we'll end up with 3 '.' delimiters in the mnemonic...).  All of this amounts to "personal preference", though.

I can certainly see the other side of this as well, as there is precedent with the 'c.' prefix used for compressed.

So count me as weakly opposed (i.e., based on a gut-feeling only).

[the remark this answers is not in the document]
aswaterman: Agreed w/Philipp.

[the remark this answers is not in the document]
kito-cheng: Maybe we could have a NOTE to mention the vendor prefix should be at least 2 char, that could prevent collision for most situation.

We (Toolchain maintainers, ISA folks, and Toolchain & runtime TG folks) are here as a gatekeeper to make sure the vendor prefix won't be too ambiguous with possible opcode.

And I guess I should express clearly: I don't have strong preference on `x.` or any other prefix, so basically I am OK if all other are OK with that, but yeah, but I feel vendor specific like `sf.`, `th.` or `vt.` prefix only is enough  and clearly, so I am also kind of weakly opposed too.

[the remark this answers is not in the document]
aswaterman: I support a two-character minimum, so that the single-character prefix space is reserved for standard purposes.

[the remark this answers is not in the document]
asb: > I support a two-character minimum, so that the single-character prefix space is reserved for standard purposes.

That sounds sensible to me.

asb:
I'd suggest replacing instances of "must" in the text with "should", as I think that aligns better with the fact these are guidelines.

ptomsich:
> I'd suggest replacing instances of "must" in the text with "should", as I think that aligns better with the fact these are guidelines.

Let's refer to RFC 2119 in the document's introduction and use SHOULD (i.e., it's a must, unless good reasons exist and are weighted) where appropriate.

In the specific instance of prefixing mnemonics, I would argue that it remains a MUST (and I would immediately put my weight behind adopting this as a binding policy for RVI members) to subdivide the mnemonic namespace.  Of course, it remains free for anyone outside of our ecosystem to not (or partially) adopt these conventions (just as they will not be bound by RVI's policies).

README.mkd
kito-cheng: Removed about encoding.

README.mkd
kito-cheng: Note added :)

kito-cheng:
Changes:
- Rebase
- Drop `encoding space` stuffs from the description in second paragraph of `Conventions for vendor extension`.
- Require vendor suffix should at least two letters
- Add note to mention there is no centralized registration is required
- Replace few `must` to `should`.

ptomsich:
@kito-cheng Thanks for updating these. Given that this addresses everyone's comments, I would suggest we get them merged now and address any further discussion topics in issues and PRs against the new HEAD.

kito-cheng:
I would like to get LGTM from LLVM community before merge since that will be followed by both LLVM and GNU community. @asb @luismarques

README.mkd
```diff
@@ -242,6 +242,63 @@ As of November 2021 the additional tail-call entry points are only
...
+
+For example:
+- `XVentanaCondOps` from Ventana
+- `Xsfcflushdlone` from SiFive
+- `Xtheade` from T-head
+
+### Assembly mnemonic
+
+In order to avoid confusion between standard extension and other vendor
+extensions, instruction mnemonics from vendor extensions must have a
```
luismarques, line 272: Here the "must" should IMO also be replaced with a "should"

README.mkd
```diff
@@ -242,6 +242,63 @@ As of November 2021 the additional tail-call entry points are only
...
+prefix corresponding to the vendor's name.
+
+e.g. `sf.` for SiFive, `vt.` for Ventana. No central registration with RISC-V
+International or elsewhere is required before the prefix is used.
+
+Vendors should also aim to follow the conventions used for naming mnemonics
+in the ratified base ISA and extensions (e.g. the use of 'w', 'd',
+'u', and 's' suffixes).
+
+The vendor suffix requires at least two letters.
```
luismarques, line 282: Isn't the prefix the one that should have two letters? E.g. `sf.magicsauce`?

ptomsich: @kito-cheng 'suffix' -> 'prefix'.

README.mkd
ptomsich: I would stick with MUST here, as this is about namespace segmentation.

Open-source projects are always free to not adopt (or selectively adopt) this convention: so they aren't bound by the MUST unless they opt in.

However, we can incorporate this section by reference into a RISC-V policy (which would then be binding—for a suitably relaxed definition of "binding"—for our members) in the future if we make this a mandatory requirement here.

README.mkd
```diff
@@ -242,6 +242,58 @@ As of November 2021 the additional tail-call entry points are only
...
+According to the RISC-V ISA spec, non-standard extensions are named using a single `x`
+followed by an alphabetical name and an optional version number.
+
+To make it easier to identify and prevent naming conflict, vendor extensions must start
+with a vendor name, which could be an abbreviation of the full name.
+
+For example:
+- `XVentanaCondOps` from Ventana
+- `Xsfcflushdlone` from SiFive
+- `Xtheade` from T-head
```
ptomsich, line 267: Can we drop the Xtheade example for now, as the names of the T-Head extensions may still change?

We are working to further subdivide these into multiple orthogonal extensions with T-Head (e.g. CMOs, instructions shadowing Zbb, instructions shadowing Zba, multiply-accumulates, etc.)...

README.mkd
kito-cheng: I didn't have strong preference on that, so I leave the`must` here.

README.mkd
kito-cheng: Yeah...that's really epic typo...

README.mkd
kito-cheng: Removed.

README.mkd
```diff
@@ -279,7 +281,7 @@ Vendors should also aim to follow the conventions used for naming mnemonics
 in the ratified base ISA and extensions (e.g. the use of 'w', 'd',
 'u', and 's' suffixes).
 
-The vendor suffix requires at least two letters.
+The vendor prefix requires at least two letters.
```
luismarques, line 284: I think this should be moved a paragraph up, to be next to the other paragraphs about the prefixes. Also, keeping with our "must" -> "should" replacements, this should probably say something like "The vendor prefix should be at least two letters long".

(BTW, I wonder if we should say "characters" instead of "letters"?)

kito-cheng:
Changes:
- Fix typo...suffix->prefix
- Remove Xtheade
- Add note to mention comply conventions isn't guarantee upstream will accept

I heard some concern from one of GNU toolchain maintainer, hope this, so I added one more note to mention upstream maintainer has final say on accepting vendor ext.

README.mkd
```diff
@@ -279,7 +281,7 @@ Vendors should also aim to follow the conventions used for naming mnemonics
...
 
 NOTE: Although no centralized registration is required, vendors should add the
```
luismarques, line 286: This is also about prefixes, so if the other paragraph is moved then this one probably should also move.

README.mkd
kito-cheng: > I think this should be moved a paragraph up, to be next to the other paragraphs about the prefixes. Also, keeping with our "must" -> "should" replacements, this should probably say something like "The vendor prefix should be at least two letters long".

Moved, that make sense to me .

> (BTW, I wonder if we should say "characters" instead of "letters"?)

Thanks for the suggestion, I don't have strong feeling on the difference between characters and letters, so I ask google translation and it give me `two letters` :P

ptomsich:
On Tue, 26 Apr 2022 at 12:26, Kito Cheng ***@***.***> wrote:

> (BTW, I wonder if we should say "characters" instead of "letters"?)
>
> Thanks for the suggestion, I don't have strong feeling on the difference
> between characters and letters, so I ask google translation and it give me two
> letters :P
>
A letter will be a printable character (as letters denote graphemes of a
writing segment) other than punctuation, diacritical marks, or numbers.
So for the Latin alphabet, the letters are [a-zA-Z].

To be precise we might restrict further to the letters of the Latin
alphabet—or to ASCII [a-z].

README.mkd
kito-cheng: > Should we also disallow [Zz] as the first letter of the vendor-prefix, as standard-extensions are names Z?

I am less worry about `Z*`, but I saw `cm.` prefixed are used in 

> Should we state somewhere that the vendor-prefix is allocated as case-insensitive?

Yeah, thanks for remind that, GNU assembly and LLVM assembly are both case-insensitive on the opcode parsing.

kito-cheng: > This is also about prefixes, so if the other paragraph is moved then this one probably should also move.

Moved

kito-cheng:
Changes:
- Tweak the order in Assembly mnemonic section
- Add note to mention vendor prefix and ext name are both case-insensitive

README.mkd
asb: FWIW I slightly prefer "should" here too. Marking anything in a "conventions" doc as a "must" seems like a logical contradiction to me. But obviously we're all coming to this with slightly different perceptions of what this doc is about!

It also feels a bit inconsistent to be heavy-handed with this rule when there are all sorts of other things vendors can do that are commonly not a good idea, but still allowable. e.g. using encoding space used by standard RISC-V extensions for their own purposes - normally a bad idea, but it's a freedom vendors have because there are times it might make sense.

But this isn't a blocking comment, it's not the biggest deal to me either way.

asb:
> I would like to get LGTM from LLVM community before merge since that will be followed by both LLVM and GNU community. @asb @luismarques

Thanks Kito for all your work bringing this together. I still have a slight preference for "should" over "must" in the one remaining instance where it's used (see inline comment), but I don't think it's a blocker.

cmuellner (approved):

PR #18, opened 2022-02-20
Update links and current statuses (batch 2022-02-20)

a4lg:
This is preparation for my upcoming proposals. Before proposals, we can clean this document to reflect latest statuses of GNU/LLVM toolchains and other external documents.

1. It updates links to external documents.
2. It reflects (not all but some of) current toolchain/ISA statuses:
    - "G" is either "IMAFD" or "IMAFD_Zicsr_Zifencei" (depending on the target ISA)
    - Because disassemblers assume availability of "Zicsr", "RV32/64GC" instead of "RV32/64 IMAFDC" would be better.
    - Because "Zicsr" and "Zmmul" extensions are at least frozen, there's no need for "fine-grained information about a target" TODO.
    - Both GNU and LLVM toolchains support parsing extension versions.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -50,10 +50,12 @@ documentation](https://gcc.gnu.org/onlinedocs/gcc/RISC-V-Options.html)
 ## Specifying the target ISA with -march
 
 The compiler and assembler both accept the `-march` flag to specify the target 
-ISA, e.g. "rv32imafd". The abbreviation "g" can be used to represent the IMAFD 
-base and extensions, e.g. `-march=rv64g`. A target `-march` which includes 
-floating point instructions implies a hardfloat calling convention, but can be 
-overridden using the `-mabi` flag (see the next section).
+ISA, e.g. "rv32imafd". The abbreviation "g" can be used to represent either 
+`IMAFD` (when targeting RISC-V ISA specification version 2.2 or earlier) or 
+`IMAFD_Zicsr_Zifencei` (version 20190608 or later) base and extensions, 
+e.g. `-march=rv64g`. A target `-march` which includes floating point 
+instructions implies a hardfloat calling convention, but can be overridden 
+using the `-mabi` flag (see the next section).
 
 The ISA subset naming conventions are described in Chapter 27 of the RISC-V 
 user-level ISA specification. However, tools do not currently follow this 
--- a/README.mkd
+++ b/README.mkd
@@ -30,7 +30,7 @@ seek input from the wider GCC and Clang developer communities for extensions
 or changes beyond the current set of command-line options supported by GCC.
 
 See the [issues 
-list](https://github.com/lowRISC/riscv-toolchain-conventions/issues) to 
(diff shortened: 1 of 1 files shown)


TommyMurphyTM1234:
>    * "G" is either "IMAFD" or "IMAFD_Zicsr_Zifencei" (depending on the target ISA)

Surely it's *only* the latter?

a4lg:
@TommyMurphyTM1234
* "IMAFD" on version 2.2 or before
* "IMAFD_Zicsr_Zifencei" on version 20190608 or later

README.mkd
```diff
@@ -30,34 +30,36 @@ seek input from the wider GCC and Clang developer communities for extensions
...
-floating point instructions implies a hardfloat calling convention, but can be 
-overridden using the `-mabi` flag (see the next section).
-
-The ISA subset naming conventions are described in Chapter 22 of the RISC-V 
+ISA, e.g. "rv32imafd". The abbreviation "g" can be used to represent either 
+`IMAFD` (when targeting RISC-V ISA specification version 2.2 or before) or 
+`IMAFD_Zicsr_Zifencei` (version 20190608 or later) base and extensions, 
+e.g. `-march=rv64g`. A target `-march` which includes floating point 
+instructions implies a hardfloat calling convention, but can be overridden 
+using the `-mabi` flag (see the next section).
```
kito-cheng, line 58: Although isn't true for GNU toolchain (`-march` implied `-mabi`), but that's out of scope for this patch, and not introduced by your change, so I think that could fixed later.

kito-cheng (approved):
LGTM.

README.mkd
```diff
@@ -30,34 +30,36 @@ seek input from the wider GCC and Clang developer communities for extensions
...
 
 The compiler and assembler both accept the `-march` flag to specify the target 
-ISA, e.g. "rv32imafd". The abbreviation "g" can be used to represent the IMAFD 
-base and extensions, e.g. `-march=rv64g`. A target `-march` which includes 
-floating point instructions implies a hardfloat calling convention, but can be 
-overridden using the `-mabi` flag (see the next section).
-
-The ISA subset naming conventions are described in Chapter 22 of the RISC-V 
+ISA, e.g. "rv32imafd". The abbreviation "g" can be used to represent either 
+`IMAFD` (when targeting RISC-V ISA specification version 2.2 or before) or
```
luismarques, line 54: nit: "or before" -> "or earlier"?

a4lg: That would work. I'm not a native speaker so this kind of feedback is always helpful.

a4lg: I reflected your comment in the document (before → earlier). Thanks!

README.mkd
cmuellner: I created an issue (#20) for this.
So we can ignore that here.

cmuellner (approved):
LGTM.
@kito-cheng any objections? If not, I will merge this within the next 2 days.

luismarques (approved):

kito-cheng:
@cmuellner OK to me, thanks :)

PR #19, opened 2022-04-20
Define prefix for T-Head vendor extensions

cmuellner:
This PR adds a commit on top of PR #17 that introduces `th` as the instruction prefix for T-Head vendor extensions.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -297,6 +297,7 @@ in the ratified base ISA and extensions (e.g. the use of 'w', 'd',
 Vendor                 | Prefix          | URL
 :--------------------- | :-------------- | :-------------
 SiFive                 | sf              | https://www.sifive.com/
+T-Head                 | th              | https://www.t-head.cn/
 Ventana Micro Systems  | vt              | https://www.ventanamicro.com/
 
 NOTE: Vendor prefixes are case-insensitive.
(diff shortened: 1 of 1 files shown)


kito-cheng:
@cmuellner thanks for this PR! I guess I should putting more effort on push #17 forward.

rjiejie:
@cmuellner Thanks for your commit,
All of instructions of T-HEAD are split to theadc & theade :)
see also https://github.com/rjiejie/XuanTie-doc

kito-cheng:
Seems like there is English version of ISA manual in https://github.com/rjiejie/XuanTie-doc, @cmuellner do you mind update the link for the PR?

rjiejie:
We will add 'th' as prefix for our ISAs and @cmuellner has made a draft version of english ISA manual, we can follow it.

cmuellner:
Sorry for not responding earlier.

Yes, we've been splitting the instructions from the SoC's user manual into RISC-V-like ISA extensions (like listed in the PR).
Along with this, we have a patchset for Binutils, QEMU, a WIP-patchset for GCC, and a document that documents the T-Head extensions in a RISC-V-like way. Some things are still in review, but once Binutils/QEMU/Doc are done, we'll send that out.

Note that the extensions don't cover every vendor-specific feature of the C906/C910.
Instead, we limit the extensions to what is relevant for kernel/userspace (e.g. no CSR bit to enable/disable unaligned memory accesses). This should make the patches much easier to review and the documentation focuses on user-exposed features (i.e. instructions that are available in S mode and U mode).

If you have questions, feel free to reach out to me via email.

kito-cheng:
Ooops, the version become date-based!?

cmuellner:
This version is autogenerated (see https://github.com/T-head-Semi/thead-extension-spec/blob/master/Makefile#L4).
So it is the date of the top commit and its commit ID.

cmuellner:
The versioning has been udpated. The document now has version 2.0.0. This PR has been updated accordingly.

kito-cheng (approved):
LGTM

kito-cheng:
More words about why it's LGTM now: I don't care about which version is used for vendor extension since I believe it's the responsibility of the vendor to maintain the meaning of version, but different version scheme is a problem because that's violate the version scheme defined the RISC-V ISA.

cmuellner:
Yes, we thought so as well.
Therefore we decided to update the versioning scheme.
Thanks for highlighting this!

PR #23, opened 2022-07-21
typo: Use plural even if we only have a single list item

cmuellner:
Since we speak of *all* extensions (even if we only have one in the
document), we need to use the plural form of the word extension.


lazyparser (commented):
LGTM. Thanks.
I'm not the maintainer of the repo, so just a comment :)

kito-cheng (approved):
LGTM

Issue #11, opened 2021-05-27
Relax the canonical order of arch string


kito-cheng:
The ISA spec has specify the ISA extension could be describe the arch string, and toolchain are using this way to control the code gen, and currently toolchain are following the order describe by the ISA spec, and emit error if order is wrong.

However we have bunch of sub-extension from several ISA extension now, zv*, zb*, zk*, zp*, zf* and zi*, and there is x*, s* and zxm* extensions, the order rule is also little bit different:
- Order between z*, s*, h*, zxm* and x* are: z* -> s* -> h* -> zxm* -> x* 
- Order rule for z*: they should be ordered first by category (second letter), then alphabetically within a category
- Order rule for h*, s*, zxm* and x*: ordered alphabetically.

Okay, so how to describe a arch has rv64gc, sfoo, zba, zfh, zicsr, zxmbar, xsf and xabc?
The canonical order is there: rv64gc_zicsr_zfh_zba_sfoo_zxmbar_xabc_xsf.
I think it's hard to write it correct without check with ISA spec.

So my proposal is relax the canonical order checking for the -march option, 

And I list few possible options:
- Relax order check for multi-letter extension, e.g. order between z*, s*, h*, zxm* and x*. [*1]
- Relax order check for whole arch string, including single letter extensions. [*1]
- Don't relax order check, but emit an error to tell user the right order.

*1: The ELF attribute emission still must in canonical order to simplify the parser logic, that's also keep the possibility of cache the checking result if we check that in OS or runtime in future.

luismarques:
Thanks for writing down this proposal Kito. I like the idea that tools that consume human-generated input would be flexible in the order they would accept, but that they would always canonicalize the ISA string. So when a tool prints something it's always in canonical order and you can easily match that string. But why should the human have to remember that order? One possible exception that was mentioned in the LLVM RISC-V sync-up call was requiring that the `i` would always come first, as that's very integral to the `rv32i` vs `rv64i` distinction, but maybe that's overthinking it. (The same logic would presumably apply to `g`).

jim-wilson:
I think the main problem is with the z and x extensions, which have unobvious orders.  The single letter extensions are pretty clear about the right ordering.  I'm Ok with relaxing the order check for user input for the multicharacter extensions.

pz9115:
[...]

Issue #1, opened 2017-10-25
Specifying custom extensions in the -march string


asb:
The string accepted by `-march` currently deviates from the ISA naming strings described in chapter 22 of the RISC-V User ISA manual. Should compilers accept strings in the form `march=rv32imcXfoobar_Xbazqux`, or a more usual `-march=rv32imc+foobar+bazqux`, or perhaps even `-march=rv32imc -mext=+foobar+bazqux` (thus separating the extension namespace from other target-specific options that might be specified in `-march`).

kito-cheng:
How about `-march=rv32imcXfoobarXbazqux`? underscore seems ugly...

For our internal porting we using `-march=rv32imcXfoobar` but we don't have more than one extension yet.

cmuellner:
Support for custom extensions was added years ago.
Therefore, closing this ticket.

PR #26, opened 2022-11-08
Define -march string

cmuellner:
This is essentially a collection of all recent change requests for the -march string:

* Define the `-march` string, as it is not exactly the RISC-V ISA naming string, and highlight the lower-case requirement (https://gcc.gnu.org/pipermail/gcc-patches/2022-November/605249.html)
* Relax the ISA string order (https://github.com/riscv-non-isa/riscv-toolchain-conventions/issues/11 and https://github.com/riscv-non-isa/riscv-toolchain-conventions/pull/14)
* Add custom extensions (https://github.com/riscv-non-isa/riscv-toolchain-conventions/issues/1)
* Add profiles support (https://lists.riscv.org/g/sig-toolchains/message/379 and https://github.com/riscv/riscv-profiles/pull/78)

Most of this patch is based on proposals and inputs from Kito and Palmer (see linked resources above).

It might not be perfect, but it could serve as a reasonable initial draft for further discussions.

Comments are welcome!

CC: @palmer-dabbelt @kito-cheng @asb @jrtc27

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -47,33 +47,149 @@ Manual](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.
 * [GCC RISC-V option 
 documentation](https://gcc.gnu.org/onlinedocs/gcc/RISC-V-Options.html)
 
-## Specifying the target ISA with -march
+## Specifying the target ISA with `-march` and `-misa-spec`
 
-The compiler and assembler both accept the `-march` flag to specify the target 
-ISA, e.g. `rv32imafd`. The abbreviation `g` can be used to represent either 
-`IMAFD` (when targeting RISC-V ISA specification version 2.2 or earlier) or 
-`IMAFD_Zicsr_Zifencei` (version 20190608 or later) base and extensions, 
-e.g. `-march=rv64g`. A target `-march` which includes floating point 
-instructions implies a hardfloat calling convention, but can be overridden 
-using the `-mabi` flag (see the next section).
+The compiler and assembler both accept the `-march` flag to specify the target
+ISA. Additionally, the `-misa-spec` flag allows specifying the ISA specification
+release, which can have a subtle difference on how the string provided to the
+`-march` flag is interpreted.
 
-The ISA subset naming conventions are described in Chapter 27 of the RISC-V 
-user-level ISA specification. However, tools do not currently follow this 
-specification (input is case sensitive, ...).
+The `-march` string can be defined using one of the following formats:
+1) using the ISA-string-based format, or
+2) using the profile-based format.
 
-If the 'C' (compressed) instruction set extension is targeted, the compiler 
+Both formats allow specifying a list of extensions that form an additive set.
+
+### ISA-string-based format
+
+The ISA-string-based format looks similar to ISA naming strings as specified
+in the ISA Extension Naming Conventions chapter of the RISC-V ISA
+specification.
+
+A valid `-march` string in the ISA-string-based format must follow these rules:
+
+* All characters of the string must be lowercase.
+* The first part of the string must be `rv32`, `rv64`, or `rv128` followed by
+  the base integer ISA (`i`, or `e`).  As `g` implies `i`, this letter can
+  also be used.
+* The next part consists of single-letter and multi-letter ISA extensions,
+  which may be in non-canonical order.
+* Single-letter ISA extensions may be separated from other single-letter extensions
+  by an underscore (`_`).
+* Multi-letter ISA extensions must be separated from other extensions
+  by an underscore (`_`).
+* Versioning of extensions follows the ISA naming string rules (e.g. `zba1p0`).
+  Single-letter extensions with versioning are still considered single-letter
+  extensions (`m1p0` is considered a single-letter extension).
+* ISA extensions can be versioned the same way as in ISA naming strings
+  (appending it directly to the ISA extension name and separate the major
+  from the minor version number using a `p`).
+  E.g. `m1p0` represents `m` in version `1.0`.
+  The selected version in case the extension's version number is not specified
+  as part of the `-march` string is compiler-specific.
+* The version separator(`p`) has a higher priority than a trailing `p` in the
+  extension name. E.g. `zba1p` will be interpreted as `zba` version 1.0 and
+  not as `zba1p`.
+
+Examples:
+
+* `rv32imac`: Valid `-march` string.
+* `rv64gc`: Valid ISA string.
+* `rv32ima_zicsr`: Valid `-march` string.
+* `rv32i2p1m2p0a2p1f2p2d2p2c_zicsr_zifencei`: Valid `-march` string.
+* `rv64gc_zba_zbb_zbc1p0`: Valid `-march` string.
+* `rv64gc_xtheadba`: Valid `-march` string.
+* `rv32i_zicsr_mafd`: Valid `-march` string.
+* `rv32i_mafd_zicsr`: Valid `-march` string.
+* `rv32i_m_a_f_d_zicsr`: Valid `-march` string.
+* `rv32i_mafdzicsr`: Invalid because `mafdzicsr` does not separate the multi-letter extension `Zicsr`.
+* `rv64gc_Zicsr`: Invalid because of the uppercase character.
+* `rv32mai`: Invalid because base ISA does not follow `rv32`.
+* `rv32i_zicsrzifencei`: Invalid because multi-letter extensions need to be
+   separated by an underscore (`zicsr` and `zifencei` are existing individual
+   extensions).
+
+NOTE: A RISC-V ISA extension might depend on other ISA extensions.
+This dependencies are honored by the `-march` string.
+E.g. `D` implies `F` and does not need to be listed explicitly.
+
+NOTE: The ISA-string-based format is a user-friendly format,
+that is different from the machine-friendly format in the ELF
+tag `Tag_RISCV_arch`.
+
+#### Interpretation versioning with `-misa-spec`
+
+The interpretation of the string provided to the `-march` flag
+in ISA-string-based format is versioned using the `-misa-spec` flag.
+The `-misa-spec` flag is incompatible with a string provided
+to the `-march` flag in the profile-based format.
+
+Possible versions are:
+
+* `2.2`
+* `20190608`
+* `20191213`
+
+Additional versions might appear in the future if a change of
+the interpretation is required.
+
+The following interpretation differences exist:
+
+* The expansion of `g` with versions before `20190608` is `imafd`.
+  The expansion of `g` with version `20190609` or later is
+  `imafd_zicsr_zifencei`.
+
+Examples:
+
+* `-march=rv64gc -misa-spec=20191213` enables RV64I and the extensions
+  `mafdc` as well as `zicsr` and `zifencei`.
+* `-march=rv64gc -misa-spec=rva22` is invalid (unknown ISA version).
+
+NOTE: The RISC-V specification is intended to be backward and
+forward-compatible. However, such a goal is usually hard to achieve and fixes
+for mistakes in specifications or misinterpretation of specifications
+may require differences in how the `-march` string is interpreted.
+The required changes are subtle and irrelevant to most users.
+However, a mechanism to enable old behavior (even if considered wrong)
+is considered a helpful solution for users that depend on this behavior.
+
+### Profile-based format
+
+The profiles-string-based format has the following form
+`-march=<profile-name>[+<option-ext>]+`.
+
+Profile names are lowercase strings of the profile name (e.g. `rva22u64`).
+It is possible to add additional extensions using the `+` character.
+
+The effect of specifying a profile is that all mandatory extensions
+of that profile will be enabled. The list of mandatory extensions
+of a profile can be found in the RISC-V profiles specification.
+
+The profiles-string-based format is implicitly versioned
+by the profile's prefix (e.g. "rva22"), therefore a change in the
+interpretation (if necessary) can be introduced by adding support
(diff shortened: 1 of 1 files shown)


cmuellner:
@palmer-dabbelt [commented the following on the GCC mailing list](https://gcc.gnu.org/pipermail/gcc-patches/2022-November/605439.html):

> IMO trying to handle this with another RISC-V spec is a waste of time:
> we've spent many years trying to follow the specs here, it's pretty
> clear they're just not meant to be read in that level of detail.  This
> sort of problem is all over the place in RISC-V land, moving to a
> different spec doesn't fix the problem.

As I wrote in my reply message, I think `-march` should be documented here (as it is already) because this repo is the common place to document things like that for GCC and LLVM.

Also, I don't think a common `-march` documentation/specification is a waste of time. But of course, I don't consider my proposal perfect. Technical comments will help to improve it. If someone else wants to drive this issue instead, feel free to do so.

README.mkd
```diff
@@ -47,34 +47,138 @@ Manual](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.
...
+* `2.2`
+* `20190608`
+* `20191213`
+
+Additional versions might appear in the future if a change of
+the interpretation is required.
+
+The following interpretation differences exist:
+
+* The expansion of `g` with versions before `20190609` is `imafd`.
```
anton-afanasyev, line 126: typo (here and below): 20190609 -> 20190608

README.mkd
```diff
@@ -47,34 +47,138 @@ Manual](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.
...
+* `rv32i2p1m2p0a2p1f2p2d2p2c_zicsr_zifencei`: Valid `-march` string.
+* `rv64gc_zba_zbb_zbc1p0`: Valid `-march` string.
+* `rv64gc_xtheadba`: Valid `-march` string.
+* `rv64gc_Zicsr`: Invalid because of the uppercase character.
+* `rv32mai`: Invalid because base ISA does not follow `rv32`.
+* `rv32i_zicsr_m`: Invalid because `m` after multi-letter extension.
+* `rv32i_zicsrzifencei`: Invalid because multi-letter extensions need to be
+   separated by an underscore (`zicsr` and `zifencei` are existing individual
+   extensions).
+
```
anton-afanasyev, line 103: Last specification defines "dependency" notion for extensions. We need to either mention that `-march=` respects this convention, or state that `-march=` doesn't follow it.
See 29.3 of 20191213: ```Some ISA extensions depend on the presence of other extensions, e.g., "D" depends on "F" and "F" depends on "Zicsr". These dependences may be implicit in the ISA name: for example, RV32IF is equivalent to RV32IFZicsr, and RV32ID is equivalent to RV32IFD and RV32IFDZicsr.```

eopXD:
Thank you for the work.

Trying to understand how the interpretation works so maybe I can help on the LLVM side. Is there an existing table for interpretation(s) to the ISA? If so this will help me clarify the confusion that I have to both "dates" and "version number" coexists (based on this PR - `2.2` and `20190608`) as an identifier to an interpretation.

README.mkd
```diff
@@ -47,34 +47,138 @@ Manual](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.
...
+A valid `-march` string in the ISA-string-based format must follow these rules:
+
+* All characters of the string must be lowercase.
+* The first part of the string must be `rv32`, `rv64`, or `rv128` followed by
+  the base integer ISA (`i`, or `e`).  As `g` implies `i`, this letter can
+  also be used.
+* The next part consists of single-letter ISA extensions, which may be in
+  non-canonical order.
+* The next part consists of multi-letter ISA extensions, which may be in
+  non-canonical order.
```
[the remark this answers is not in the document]
cmuellner: Will do.

README.mkd
```diff
@@ -47,34 +47,138 @@ Manual](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.
...
+
+* `rv32imac`: Valid `-march` string.
+* `rv64gc`: Valid ISA string.
+* `rv32ima_zicsr`: Valid `-march` string.
+* `rv32i2p1m2p0a2p1f2p2d2p2c_zicsr_zifencei`: Valid `-march` string.
+* `rv64gc_zba_zbb_zbc1p0`: Valid `-march` string.
+* `rv64gc_xtheadba`: Valid `-march` string.
+* `rv64gc_Zicsr`: Invalid because of the uppercase character.
+* `rv32mai`: Invalid because base ISA does not follow `rv32`.
+* `rv32i_zicsr_m`: Invalid because `m` after multi-letter extension.
```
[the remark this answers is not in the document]
cmuellner: Will do.

README.mkd
cmuellner: Thx!

README.mkd
cmuellner: Thanks for highlighting this.
I think it is expected by users that the march string follows this implicit enablement of extensions.
I will add a sentence to clarify this.

cmuellner:
> Trying to understand how the interpretation works so maybe I can help on the LLVM side. Is there an existing table for interpretation(s) to the ISA? If so this will help me clarify the confusion that I have to both "dates" and "version number" coexists (based on this PR - `2.2` and `20190608`) as an identifier to an interpretation.

I tried to cover exactly this by the amount of examples included in the PR.
If there are unclear cases, then I think we should list them in this document anyway.

Regarding the ISA versions: we will just have to support all existing versioning schemes: so versions are well-ordered, but not lexicographic sortable, strings.

cmuellner:
I've updated the PR according to the provided feedback.
Thanks for the comments!

cmuellner:
Kito mentioned that there should be more details regarding the versioning of extensions. I've added some details about that to the PR.

jrtc27:
We should have -march=isa+ext for completeness (yes, _ works in that case, but then you have to care about whether it’s an ISA or profile as your “base”). We also need -march=+foo so you can add to whatever the default is (whether that’s from a flag in user-provided CFLAGS or the system default when not specified). Maybe also -march=-foo, though that gets a bit weird (but that already needs to work in the assembler due to .option arch, so give it the same semantics).

jrtc27:
I also worry about what happens if a user gives CFLAGS=“-march=isa -misa-spec=…” and a build system then adds -march=profile given you’ve said it’s an error to combine those.

README.mkd
```diff
@@ -47,34 +47,150 @@ Manual](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.
...
+
+* All characters of the string must be lowercase.
+* The first part of the string must be `rv32`, `rv64`, or `rv128` followed by
+  the base integer ISA (`i`, or `e`).  As `g` implies `i`, this letter can
+  also be used.
+* The next part consists of single-letter and multi-letter ISA extensions,
+  which may be in non-canonical order.
+* Single-letter ISA extensions may be separated from other single-letter extensions
+  by an underscore (`_`).
+* Multi-letter ISA extensions must be separated from other extensions
```
nick-knight, line 79: Vol. II Ch. 29 says:

> Extensions with the “Z” prefix must be separated from other multi-letter extensions by an under-
score, e.g., “RV32IMACZicsr Zifencei”.

Note there is no underscore between `C` and `Zicsr`.

README.mkd
```diff
@@ -47,34 +47,150 @@ Manual](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.
...
+
+### ISA-string-based format
+
+The ISA-string-based format looks similar to ISA naming strings as specified
+in the ISA Extension Naming Conventions chapter of the RISC-V ISA
+specification.
+
+A valid `-march` string in the ISA-string-based format must follow these rules:
+
+* All characters of the string must be lowercase.
```
nick-knight, line 71: Vol. II Ch. 29 says:

> The ISA naming strings are case insensitive.

Indeed, the examples given therein use mixed case.

nick-knight (commented):
I support this effort. But I think it is worth calling out explicitly where it breaks compatibility with the ISA string as defined in Vol. II Ch. 29, since existing software and build systems are following those conventions already for `march`. I give two examples below of compatibility breaks. These are just the first two that immediately jumped out at me (I didn't go through the whole PR carefully).

asb:
I do worry that profile strings might be rather cryptic and easily confusable (by humans) with ISA naming strings. If it's agreed this is problematic, it's not clear if it's best handled in the profiles spec or in how the profile names are represented in `-march` arguments.

`i2020u32` is cryptic but perhaps less easily confusable than `rvi20u32`. As we get closer to 2032, `rvi32u32` vs the ISA naming string `rv32i...` seems potentially confusing. As Andrew argues in https://github.com/riscv/riscv-profiles/issues/82 it's possible that the confusability won't matter as almost nobody is using ISA naming strings.

cmuellner:
As mentioned in the profiles ticket, would it be ok, to mention in a note that tools/compilers may impose restrictions about which march string format is accepted (e.g. via prefix or cmdln flag) and that users may look in to the compiler manual for details? So we would just define the formats here and leave some room for further restrictions for tools/compilers (if any).

cmuellner:
> We should have -march=isa+ext for completeness (yes, _ works in that case, but then you have to care about whether it’s an ISA or profile as your “base”). We also need -march=+foo so you can add to whatever the default is (whether that’s from a flag in user-provided CFLAGS or the system default when not specified). Maybe also -march=-foo, though that gets a bit weird (but that already needs to work in the assembler due to .option arch, so give it the same semantics).

I'm not sure if we should define this here (unless enough people want this). The idea is good, but it feels a bit like an added-value feature of a tool/compiler.

> I also worry about what happens if a user gives CFLAGS=“-march=isa -misa-spec=…” and a build system then adds -march=profile given you’ve said it’s an error to combine those.

Same for this. I would let this behavior be defined by the tool as I fear that different tools prefer different behavior (e.g. accepting the last provided flag vs. throwing an error vs. ./configure flag to define the policy...).

I would prefer to restrict the description to a minimum that's usable by a user. Compilers and tools can then derive from here and extend more features.

Would this be ok?

README.mkd
cmuellner: I properly described what I wanted to express, but I was not aware that `-march=rv64imafdczbb` is valid in GCC and clang. I will update this to match the currently implemented behavior to reduce the differences to the ISA strings.

Fun fact: GCC currently accepts `-march=rv64imafdczbbzbazifencei` and clang does not.
But I don't think that's an issue.

README.mkd
cmuellner: This is listed as the first bullet "All characters of the string must be lowercase".
So I think this is already properly documented.

jrtc27:
> > We should have -march=isa+ext for completeness (yes, _ works in that case, but then you have to care about whether it’s an ISA or profile as your “base”). We also need -march=+foo so you can add to whatever the default is (whether that’s from a flag in user-provided CFLAGS or the system default when not specified). Maybe also -march=-foo, though that gets a bit weird (but that already needs to work in the assembler due to .option arch, so give it the same semantics).
> 
> I'm not sure if we should define this here (unless enough people want this). The idea is good, but it feels a bit like an added-value feature of a tool/compiler.
> 
> > I also worry about what happens if a user gives CFLAGS=“-march=isa -misa-spec=…” and a build system then adds -march=profile given you’ve said it’s an error to combine those.
> 
> Same for this. I would let this behavior be defined by the tool as I fear that different tools prefer different behavior (e.g. accepting the last provided flag vs. throwing an error vs. ./configure flag to define the policy...).
> 
> I would prefer to restrict the description to a minimum that's usable by a user. Compilers and tools can then derive from here and extend more features.
> 
> Would this be ok?

If GCC and LLVM do different things for any of these it will be a pain point, so no, I think they really do need to be documented conventions. Otherwise either nobody will use them or they'll use them and be locked into one compiler for no good reason.

README.mkd
nick-knight: I realize now that this is already enforced by existing toolchains, so in this case I was wrong to say that

> existing software and build systems are following those conventions already for `march`.

I still think it is worth a remark about incompatibilities with the ISA spec, but such nonnormative text can always be added later.

kito-cheng:
Following content has send in another private mail thread, but I think it could be also post here:

---
I agree the naming of profile can be improved to prevent the confusion, but actually I am less concern about the confusion on the `-march` string in practice.

Consider the following two cases (and use the most confused possible
name rvi32u32):
1. -march= with profile name only without extra extension
   -march=rvi32u32  - the possible and legal misspell version should
be rv32iu32 which is rv32i base plus U extension with version 32,
that's generally a useless configuration, should got error soon when
trying to build/compile/assemble anything.

2. -march= with profile name only with extra extensions
  -march=rvi32u32+xsffoo profile rvi32u32 with xsffoo vendor
extension, and that's apparently a profile based arch string, since
there is a plus sign there, it's only used in profile style -march
string
  Arguable point might has 2:
    first is plus sign not included in the ISA string now, but
used/defined in future
      - Low probability, but I guess we could...try to stop when that happened.
    second is we might also want to accept  the plus sign on the
normal ISA-string-like -march option.
      - This could be prevent when we relax the canonical order
requirement of -march string


So based on the above two situations I think that should not be
confusing (or misuse with misspellings) in practice.

cmuellner:
In today's Toolchain SIG call, we agreed to split this PR into multiple ones not to mix different issues and to move this forward:
1. document current -march string
2. document -march string relaxation
3. add custom extensions
4. add profiles support

cmuellner:
Profiles support landed in #36.
Other changes in the PR did not reach consensus.
This PR attempted to do too much.
Therefore, I'm closing this.

PR #27, opened 2022-12-06
Vendor extensions: Add 'XTheadVdot' T-Head vendor extensions

rjiejie:
This patch defines the T-Head vendor extensions (XTheadVdot), which is documented here:
  https://github.com/T-head-Semi/thead-extension-spec

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -307,15 +307,17 @@ NOTE: Vendor prefixes are case-insensitive.
 Vendor  | Name            | Version        | ISA Document
 :------ | :-------------- | :------------- | :---------------
 T-Head  | XTheadCmo       | 2.0.0          | [T-Head ISA extension specification](https://github.com/T-head-Semi/thead-extension-spec/releases/download/2.0.0/xthead-2022-09-05-2.0.0.pdf)
-T-Head  | XTheadSync      | 2.0.0          | [T-Head ISA extension specification](https://github.com/T-head-Semi/thead-extension-spec/releases/download/2.0.0/xthead-2022-09-05-2.0.0.pdf)
 T-Head  | XTheadBa        | 2.0.0          | [T-Head ISA extension specification](https://github.com/T-head-Semi/thead-extension-spec/releases/download/2.0.0/xthead-2022-09-05-2.0.0.pdf)
 T-Head  | XTheadBb        | 2.0.0          | [T-Head ISA extension specification](https://github.com/T-head-Semi/thead-extension-spec/releases/download/2.0.0/xthead-2022-09-05-2.0.0.pdf)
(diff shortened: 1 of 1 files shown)


README.mkd
```diff
@@ -318,6 +318,7 @@ T-Head  | XTheadMac       | 2.0.0          | [T-Head ISA extension specification
 T-Head  | XTheadMemPair   | 2.0.0          | [T-Head ISA extension specification](https://github.com/T-head-Semi/thead-extension-spec/releases/download/2.0.0/xthead-2022-09-05-2.0.0.pdf)
 T-Head  | XTheadMemIdx    | 2.0.0          | [T-Head ISA extension specification](https://github.com/T-head-Semi/thead-extension-spec/releases/download/2.0.0/xthead-2022-09-05-2.0.0.pdf)
 T-Head  | XTheadSync      | 2.0.0          | [T-Head ISA extension specification](https://github.com/T-head-Semi/thead-extension-spec/releases/download/2.0.0/xthead-2022-09-05-2.0.0.pdf)
+T-Head  | XTheadVdot      | 2.2.0          | [T-Head ISA extension specification](https://github.com/T-head-Semi/thead-extension-spec/releases/download/2.2.0/xthead-2022-12-04-2.2.0.pdf)
```
kito-cheng, line 321: I don't really mind the version scheme for vendor extension, but 2.2 seems like spec version rather than extension version, and I don't see individual extension version for other T-Head extensions, does it means all other T-Head extensions also bump to 2.2?

rjiejie: @kito-cheng From my personal opinion, we should set version of all T-Head ISAs as v1.0, and have individual extension version.
here I copy from @cmuellner's scheme just only, maybe @cmuellner can clarify this, he is our drafter of open-source version scheme :)

cmuellner: The version listed here is the version of the document in which the extension is the first time documented.
Since the extensions are stable (in the sense of being implemented in hardware, not in the weaker RVI sense), their version is irrelevant (similar to the version of ratified standard extensions).

I chose the document version to avoid confusion with a "1.0" version that does not match the referenced document. But I don't mind if we switch to "1.0" for all extensions.

For me, it is also ok to change all links to the latest spec version.

cmuellner:
This has been resolved in the specification: we have introduced versioning on a per-extension level and all extensions got version 1.0.

I've updated the table accordingly: https://github.com/riscv-non-isa/riscv-toolchain-conventions/pull/28

@rjiejie: please update this PR accordingly (use version 1.0).

README.mkd
cmuellner: The table was updated in #28.
So once this PR is updated, this can get merged.

PR #29, opened 2023-01-23
Add vendor specific information for OpenHW Group CORE-V processors

jeremybennett:
Support for all OpenHW CORE-V processors is planned to be
	upstreamed to both GCC and Clang/LLVM

	* README.mkd (List of vendor prefixes): Add CV for all OpenHW Group CORE-V cores. (List of vendor extensions) Add all eight CORE-V ISA extensions.

Signed-off-by: Jeremy Bennett <jeremy.bennett@embecosm.com>

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -296,16 +296,27 @@ in the ratified base ISA and extensions (e.g. the use of 'w', 'd',
 
 Vendor                 | Prefix          | URL
 :--------------------- | :-------------- | :-------------
+Open Hardware Group    | cv              | https://www.openhwgroup.org/
 SiFive                 | sf              | https://www.sifive.com/
 T-Head                 | th              | https://www.t-head.cn/
 Ventana Micro Systems  | vt              | https://www.ventanamicro.com/
 
 NOTE: Vendor prefixes are case-insensitive.
 
+NOTE: OpenHW cores are all branded as CORE-V, hence the prefix.
+
 ### List of vendor extensions
 
 Vendor  | Name            | Version        | ISA Document
 :------ | :-------------- | :------------- | :---------------
+OpenHW  | Xcvalu          | 1.0.0          | [CORE-V Instruction Set Extensions](https://github.com/openhwgroup/cv32e40p/blob/dev/docs/source/instruction_set_extensions.rst)
+OpenHW  | Xcvbi           | 1.0.0          | [CORE-V Instruction Set Extensions](https://github.com/openhwgroup/cv32e40p/blob/dev/docs/source/instruction_set_extensions.rst)
+OpenHW  | Xcvbitmanip     | 1.0.0          | [CORE-V Instruction Set Extensions](https://github.com/openhwgroup/cv32e40p/blob/dev/docs/source/instruction_set_extensions.rst)
+OpenHW  | Xcvelw          | 1.0.0          | [CORE-V Instruction Set Extensions](https://github.com/openhwgroup/cv32e40p/blob/dev/docs/source/instruction_set_extensions.rst)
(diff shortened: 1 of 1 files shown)


README.mkd
```diff
@@ -296,16 +296,27 @@ in the ratified base ISA and extensions (e.g. the use of 'w', 'd',
...
 
 NOTE: Vendor prefixes are case-insensitive.
 
+NOTE: OpenHW cores are all branded as CORE-V, hence the prefix.
+
 ### List of vendor extensions
 
 Vendor  | Name            | Version        | ISA Document
 :------ | :-------------- | :------------- | :---------------
+OpenHW  | Xcvalu          | 2.0.0          | [CORE-V Instruction Set Extensions](https://docs.openhwgroup.org/projects/cv32e40p-user-manual/en/latest/instruction_set_extensions.html)
```
cmuellner, line 312: I could not find any reference to `Xcvalu` or the other extension names on the linked page (even when searching for these extension names using the search box on the top left).

Also, I cannot see how the specified version `2.0.0` correlates to the contents of the linked documentation.

jeremybennett: @cmuellner Ah - I linked to the main architecture description of the architectural extensions. There are separate documents  describing the `-march` naming convention and listing the builtin functions for those extensions.  Rather than make the table very large, should I provide those links generically for CORE-V at the bottom of the table? Thanks.

cmuellner: A link to a document where the extensions are defined (name, version and included instructions) would be sufficient. I.e. the extension name `xcvalu` and its version should not be defined in the toolchain conventions but in the (referenced) extension documentation.

jeremybennett:
@cmuellner It seemed better to fix the CORE-V documentation to make the ISA extension naming explicit in the architecture document, which has now been reviewed and committed. I have updated the links to point to this updated document.

I have corrected the version number - I misunderstood it as the version of the processor (which is version 2), not the ISA extensions (which are version 1).  I have add supplementary documentation at the bottom of the table.

Patch rebased, so should be clean to merge.

kito-cheng:
This might not very related to this PR, but I am really curious about that: `RISC-V international plans to remove versioning of ISA extensions from the new standard` (from this [doc](https://github.com/openhwgroup/core-v-sw/blob/master/specifications/corev-isa-extension-naming.md)), does here any discussion for this, mailing list or somewhere?

jeremybennett:
@kito-cheng It's relevant, because we saw that announcement, which is why we have gone for a single version of the ISA extensions being supported in the tool chain.  Although the ISA extensions are new to CV32E40Pv2, they are derived from the ETHZ/Unibo PULP work, but with different instruction encodings.  We could have tried to support the PULP encodings as version 1 and the CORE-V encodings as version 2.  Instead we are just supporting the CORE-V encodings as a single version.

In any case, in practice multiple versions is hard for GNU tool chains - the GNU assembler doesn't really have the complete infrastructure to make it work (we tried).

cmuellner (approved):

cmuellner:
LGTM

@kito-cheng, any objections to merging this?
If I hear nothing back by the end of the week, I will merge this PR.

kito-cheng (commented):
Although I am little bit concern about the extension version isn't write down in the linked document (only document version), but no strong opinion on that since it's vendor extension, and version are managed by vendor.

kito-cheng:
So treat my comment as weak LGTM

jeremybennett:
Thanks @kito-cheng I'll get the linked document updated to make explicit that all extensions are only ever a single version.

pz9115 (approved):
LGTM

jeremybennett:
Many thanks. J

PR #31, opened 2023-01-31
Vendor extensions: Fix alphabetical order

cmuellner:
A recent patch introduced the SiFive extensions at the wrong position. Let's fix that.

Signed-off-by: Christoph Müllner <christoph.muellner@vrull.eu>


kito-cheng (approved):

kito-cheng:
Thanks for fix my stupid error

PR #34, opened 2023-04-04
Update authorship information

asb:
My listed email address is long out of date, and this document has gained a range of additional contributors since then. Rather than having an out of date list, I suggest just linking to the history on github for people who want to view contributors.


cmuellner (approved):
LGTM

PR #38, opened 2023-07-12
Modified TODO chapter content

yulong18:
I deleted these options because I found it already done.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -346,10 +346,6 @@ specification.
 
 ## TODO
 
-* `-mdiv`, `-mno-div`, `-mfdiv`, `-mno-fdiv`, `-msave-restore`, 
-  `-mno-save-restore`, `-mstrict-align`, `-mno-strict-align`, 
-  `-mexplicit-relocs`, `-mno-explicit-relocs`
-
 ## Appendix: Exposing a vendor-specific extension across the toolchain
 
 TODO.


asb:
Getting rid of the TODO section and moving anything remaining to issues makes sense to me. I'm not sure all the listed items are covered though - e.g. explicit-relocs, -mdiv etc.

yulong-plct:
Yes, I'm sure. You can see code in riscv.opt and know about these options in the [manual](https://gcc.gnu.org/onlinedocs/gcc/RISC-V-Options.html). @asb

asb:
Right, but this repo is meant to cover documentation for options that are shared between GCC and Clang. I'm not fundamentally opposed to pointing to documentation for clang or GCC for these options, but we should do this in a case-by-case basis, and it's probably better overall to just write something reasonable in this doc.

cmuellner:
All this PR needs to land is also to delete the two lines above the removed text.

apazos:
We discussed this PR at Toolchain SIG meeting on 04/10 and decided to remove the TODO chapter and open issues for the features not implemented. I have pushed the PR #82 that removes the TODO chapter.

apazos:
@asb, can you please open LLVM bugs to track / add support to -mfdiv, -mno-fdiv, -mexplicit-relocs, -mno-exlicit-relocs, -mdiv
-mno-div for parity with GCC. I have removed the TODOs from the this document.

cmuellner (approved):
LGTM (given we have a follow-up PR for the section header).

cmuellner:
This PR can't be merged because of conflicts.
The change is also part of #82. Therefore, closing this.

Thanks for providing the PR!

PR #43, opened 2024-03-11
Add rule for vendor CSR naming scheme

kito-cheng:
In the last LLVM sync up meeting, we have dicussed the vendor CSR stuffs, and we realized that we have rule for vendor extenison name and assembly mnemonic, but no rule for vendor CSR naming scheme yet.

So it's would be great we have one before anyone start to upstream their own CSR definition.

The naming scheme basically follow the assembly mnemonic - use vendor prefix, the only difference is use underscore rather than dot, because it could be used in marco definition, but dot can't, that's an important character since that allow user to use marco before assmbler support, and eaiser to migrate to new version once assmbler support.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -322,6 +322,15 @@ Vendors should also aim to follow the conventions used for naming mnemonics
 in the ratified base ISA and extensions (e.g. the use of 'w', 'd',
 'u', and 's' suffixes).
 
+### CSR naming scheme
+
+Vendors may define their own CSRs within the custom read-only CSR address range
+specified in the RISC-V ISA spec. However, to avoid conflicts, each vendor CSR
+must include a prefix corresponding to the vendor's name.
+
+The vendor prefix should match the prefix defined in assembly mnemonics and be
+separated by a dot, e.g., `th.vxrm`.
+
 ### List of vendor prefixes
 
 


kito-cheng:
cc @asb, @jeremybennett, @topperc, @fanghuaqi, @apazos, @JeffreyALaw

README.mkd
```diff
@@ -291,6 +291,20 @@ Vendors should also aim to follow the conventions used for naming mnemonics
 in the ratified base ISA and extensions (e.g. the use of 'w', 'd',
 'u', and 's' suffixes).
 
+### CSR naming scheme
+
+Vendors may define their own CSRs within the custom read-only CSR address range
+specified in the RISC-V ISA spec. However, to avoid conflicts, each vendor CSR
+must include a prefix corresponding to the vendor's name.
+
+The vendor prefix should match the prefix defined in assembly mnemonics. The
```
fanghuaqi, line 300: Looks good to me. Are there any rules for CSR addressing?

kito-cheng: ISA spec has define custom/vendor CSR addressing range[1], but I don't really care the CSR is defined within the range or not...since that's vendor's issue rather than open source toolchain maintainer's issue.

[1] https://github.com/riscv/riscv-isa-manual/blob/main/src/priv-csrs.adoc#csr-listing

apazos: Thanks for updating the specs, @kito-cheng!

cmuellner:
Binutils already has vendor CSRs for XTheadVector: https://github.com/bminor/binutils-gdb/blob/master/include/opcode/riscv-opc.h#L4556

There, the vendor prefix with the dot is used.

I did not get the limitation that the dot would cause, is there an example?

cmuellner:
Here is an example of how to use the vendor prefix with dot in C macros to read CSRs:

```c
#define __stringify_1(x...)   #x
#define __stringify(x...)     __stringify_1(x)

#define VENDOR_PREFIX_TH th.

#ifdef USE_TH
#define PREFIX VENDOR_PREFIX_TH
#else
#define PREFIX
#endif

unsigned long read_vstart(void)
{
        register unsigned long v;
        __asm__ __volatile__("csrr %0, " __stringify(PREFIX) "vstart" : "=r" (v) : : "memory" );
        return v;
}
```

kito-cheng:
Oh, I didn't notice thead vector already added some vendor CSR (even I have take few time to review T-head vector stuff in GCC :P), I also prefer kinda prefer using dot as separator before, however Craig point out using underscore rather than dot, because it could be used for .S file, give some example:

foo.S:
```asm
#if BINUTILS_OR_LLVM_TOO_OLD
#define th_vstart 100
#endif

bar:
    csrr a0, th_vstart
    ret
```

But that is not work when it use dot as separator:

foo.S:
```asm
#if BINUTILS_OR_LLVM_TOO_OLD
#define th.vstart 100
#endif

bar:
    csrr a0, th.vstart
    ret
```

cmuellner:
The T-Head support in Binutils was merged on Nov 23, 2023.
So, it is part of the last Binutils release (2.42).

I prefer that people use recent releases instead of building workarounds to use old tools, so my excitement about supporting such a workaround is limited (if you want new features, then use new tools). Further, I prefer the dot because it is equal to what we do with vendor instructions, and this consistency is, for me, much more important than enabling a workaround mechanism for not having to update tools.

Anyway, I don't have the last word here, so here is a possible solution:
What can be done (besides simply using the dot as the separator) is to specify underscore as the separator and keep support for the six `th.`-CSRs in parallel. We can request to rename the CSRs in the XTheadVector spec, so that the situation is somehow contained. However, I don't want to drop support for the six `th.`-CSRs because this would clearly break existing code.

kito-cheng:
@cmuellner I agree with your point, also I think it's important thing is we missed the T-head CSR has merged into binutils during last LLVM sync up discussion...I am not intend to make T-head CSR become special case :)

Let me raise the discussion in next LLVM sync meeting again.

kito-cheng:
Oh shit...I just forgot to add anything and push nothing...

git meme

cmuellner (approved):
LGTM

Issue #42, opened 2024-03-08
medlow code generation


davidharrishmc:
I am trying to understand and explain mcmodel=medlow behavior.  I see that it is supposed to generate addresses with lui/addi and that there is a proposal to deprecate.

I tried comparing -mcmodel= medlow and medany with a code snippet for la.
```
.section .text.init
.globl rvtest_entry_point
rvtest_entry_point:
    la t0, Y
    lw t1, 0(t0)
 
.align 20
X:
.word 0x00000000
Y:
.word 0x00000000
```
The assembler produced the same code for both for -march=rv64gc.

```
    80000000:	00100297          	auipc	t0,0x100
    80000004:	00428293          	add	t0,t0,4 # 80100004 <Y>
    80000008:	0002a303          	lw	t1,0(t0)
```

I would have expected medlow to generate
lui t0, 0x80100
addi t0, t0, 0x4

Can you tell me if medlow is becoming the same as medany or what is happening here?

Thank you!

TommyMurphyTM1234:
> I am trying to understand and explain mcmodel=medlow behavior. 

Does this help at all?

* https://www.sifive.com/blog/all-aboard-part-4-risc-v-code-models

> there is a proposal to deprecate.

Where is this proposal?

davidharrishmc:
@TommyMurphyTM1234 

I have looked at these references but am wondering if they are outdated.

The scanty proposal is on this repo:
https://github.com/riscv-non-isa/riscv-toolchain-conventions

'''
Issues for consideration

It has been proposed to deprecate the medlow code model and rename medany to medium.
'''

Also, on this repo, medlow is supposed to use lui, while medany is supposed to use auipc.  In my experiment with a recent GCC, both produce auipc for la.  I'm wondering if medlow is now handled like medany in all cases, or if there's something funny about my experiment.

'''
-mcmodel=medlow. The program and its statically defined symbols must lie within a single 2GiB address range, between the absolute addresses -2GiB and +2GiB. lui and addi pairs are used to generate addresses.

-mcmodel=medany. The program and its statically defined symbols must lie within a single 4GiB address range. auipc and addi pairs are used to generate addresses.

'''

TommyMurphyTM1234:
> I have looked at these references but am wondering if they are outdated.

On what basis are you wondering that?

Maybe @kito-cheng or @cmuellner might be able to comment on this issue?

cmuellner:
[...]

PR #44, opened 2024-03-12
Remove hint for deprecating medlow

cmuellner:
A hint that medlow might get deprecated in the future is a) useless, and b) misleading.
Let's drop that.

This came up as part of https://github.com/riscv-non-isa/riscv-toolchain-conventions/issues/42.


kito-cheng (approved):
Ooops, I am surprise that, anyway I never thinking to doing that, also the timing to doing that is already gone for long time.

PR #46, opened 2024-04-12
Convert to adoc format

kito-cheng:
The final goal is convert this repo into the RVI document format, but I think it's at least a good start to convert markdown to adoc format.

Convert by [kramdown](https://github.com/asciidoctor/kramdown-asciidoc)


cmuellner:
Thanks again for creating this PR!

As we went ahead with #52, let's close this one.

PR #47, opened 2024-05-06
Adding mechanism to amend to the target ISA string via -march

cmuellner:
This has been identified to be helpful when building the same sources multiple times for different ISA targets (e.g. without Zbb and with Zbb).

This came up in a discussion about a patchset that introduces ifunc routines for glibc. See also:
* https://sourceware.org/pipermail/libc-alpha/2024-April/156204.html (patchset)
* https://sourceware.org/pipermail/libc-alpha/2024-April/156412.html (Palmer mentions `-march=+zbb`)
* https://sourceware.org/pipermail/libc-alpha/2024-May/156553.html (my response that also links to here)

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -101,6 +101,27 @@ NOTE: Any output of ISA string like `Tag_RISCV_arch` must be canonical order.
 NOTE: Cross-tool argument are highly recommended passed in canonical order for
       backward compatible.
 
+### Amending extensions with -march
+
+The compiler and assembler accept the `-march` flag to append extensions
+to the target ISA by assigning a new extension (or an extension list) via `=+`.
+I.e., `-march=rv64gc -march=+zba_zbb` is equivalent to `-march=rv64gc_zba_zbb`.
+
+The rules for the amendment string are:
+
+1. Bases cannot be part of an amended string.
+2. Extensions must not be in canonical order.
+3. Multi-letter extensions must be separated by underscores.
+4. Multiple amendments are allowed.
+5. The resulting ISA string is derived by the concatenation of the target
+   ISA with all amendment strings using an underscore (in the order that
+   is provided on the command line).
+
+Examples:
+
+* `-march=rv64gc -march=+zba_zbb` is equivalent to `-march=rv64gc_zba_zbb`
+* `-march=rv32i -march=+gcv -march=+zbb` is equivalent to `-march=rv32gcv_zbb`
+
 ### Issues for consideration
 * Whether `riscv32` and `riscv64` should be accepted as synonyms for `rv32` 
 and `rv64`.


cmuellner:
Ping @kito-cheng @asb @palmer-dabbelt @JeffreyALaw @preames @topperc

kito-cheng (commented):
I don't really like this, this will made the `-march` option become context-aware and not easy to find the full arch string, also here is few...interesting case we need figure out:

1. `-march=rv64gc -march=+zba -march=rv64gc_zbb`: Should be `rv64gc_zba` or `rv64gc_zbb`?
2. `-mcpu=sifive-x280 -march=+zbc`: `-mcpu` will implied `-march` if `-march` is not given, and arch string of `sifive-x280` is `rv64imafdcv_zfh_zba_zbb_zvfh_zvl512b`, so it should be `-march=rv64imafdcv_zfh_zba_zbb_zbc_zvfh_zvl512b`.
3. `-march=+zba`: No full arch string is given, only `-march=+`, it should interact with the default `-march` string from compiler?
4. `-march=+zba -march=rv64gc_zbb`: should be `rv64gc_zba_zbb`?

TommyMurphyTM1234:
> This has been identified to be helpful when building the same sources multiple times for different ISA targets (e.g. without Zbb and with Zbb).

Is there a link to any previous related discussion?

On the face of it I agree with @kito-cheng that this change seems to make things overcomplicated and possibly ambiguous in certain cases.

cmuellner:
> 1. `-march=rv64gc -march=+zba -march=rv64gc_zbb`: Should be `rv64gc_zba` or `rv64gc_zbb`?

I could not find any documentation that multiple `-march=...` arguments are allowed right now.
So, that's undefined behavior. We could disallow multiple `-march=...` arguments if `-march=+` is used
("ambiguous target ISA").

> 2. `-mcpu=sifive-x280 -march=+zbc`: `-mcpu` will implied `-march` if `-march` is not given, and arch string of `sifive-x280` is `rv64imafdcv_zfh_zba_zbb_zvfh_zvl512b`, so it should be `-march=rv64imafdcv_zfh_zba_zbb_zbc_zvfh_zvl512b`.

Yes, this should be supported.

> 3. `-march=+zba`: No full arch string is given, only `-march=+`, it should interact with the default `-march` string from compiler?

Yes, if no base is provided, then the configured base should be used.

> 4. `-march=+zba -march=rv64gc_zbb`: should be `rv64gc_zba_zbb`?

For a single `-march=...` I think this would be fine.

I've updated the PR to address the mentioned cases.

palmer-dabbelt:
> > 1. `-march=rv64gc -march=+zba -march=rv64gc_zbb`: Should be `rv64gc_zba` or `rv64gc_zbb`?
> 
> I could not find any documentation that multiple `-march=...` arguments are allowed right now. So, that's undefined behavior. We could disallow multiple `-march=...` arguments if `-march=+` is used ("ambiguous target ISA").

Please stop doing this.  Just because you guys unilaterally decided to make a document that doesn't match what existing toolchains do doesn't mean that we've got undefined behavior.  Continuing to pretend that what's written here has any bearing on reality just risks confusing users.  Everyone else has figured out how to work upstream.

cmuellner:
> > > 1. `-march=rv64gc -march=+zba -march=rv64gc_zbb`: Should be `rv64gc_zba` or `rv64gc_zbb`?
> > 
> > 
> > I could not find any documentation that multiple `-march=...` arguments are allowed right now. So, that's undefined behavior. We could disallow multiple `-march=...` arguments if `-march=+` is used ("ambiguous target ISA").
> 
> Please stop doing this. Just because you guys unilaterally decided to make a document that doesn't match what existing toolchains do doesn't mean that we've got undefined behavior. Continuing to pretend that what's written here has any bearing on reality just risks confusing users. Everyone else has figured out how to work upstream.

Consider focusing on the (indeed valid) technical arguments and avoid personal attacks.

PR #49, opened 2024-05-29
Add document for -mstrict-align and -mno-strict-align

kito-cheng:
We have those two option for a while, but not document in this repo yet, however we are intend to claify the scope of the option recently, so it's time to document down what we have agree among different community.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -398,11 +398,34 @@ specification, and in the draft [CORE-V Builtin
 Function](https://github.com/openhwgroup/core-v-sw/blob/master/specifications/corev-builtin-spec.md)
 specification.
 
+## Common Toolchain Command Line Options
+
+This section lists common RISC-V specific toolchain command line options.
+
+### `-mstrict-align`/`-mno-strict-align`
+Indicates that the compiler should not assume that unaligned scalar and
+unaligned vector memory references are handled by the system.
+
+`-mstrict-align`: The compiler disallows misaligned memory access.
+`-mno-strict-align`: The compiler allows misaligned memory access.
+
+The compiler's behavior will follow this order of precedence:
+
+- Use the setting from `-mstrict-align` / `-mno-strict-align` if either option
+  is given, taking the last one specified.
+- Use the setting from `-mtune` if `-mstrict-align` / `-mno-strict-align` is not given.
+- Use the setting from `-mcpu` if neither of the above options is given.
+- Use the compiler's default setting if none of the above options are provided.
+
+NOTE: Non-strict also known as unaligned access or misaligned access
+NOTE: The compiler may generate misaligned access if the program violates the
+      alignment assumption.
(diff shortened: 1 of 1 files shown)


kito-cheng:
Ref: 
Discussion in riscv-c-api-doc: https://github.com/riscv-non-isa/riscv-c-api-doc/issues/73
GCC patch: https://patchwork.sourceware.org/project/gcc/patch/cc653588-ac19-4f09-9693-78a4d47cfb1f@gmail.com/

kito-cheng:
CC: @nick-knight @kito-cheng @vineetgarc @asb @aswaterman

topperc:
CC @preames

README.mkd
```diff
@@ -400,11 +400,22 @@ specification, and in the draft [CORE-V Builtin
...
 specification.
 
+## Common Toolchain Command Line Options
+
+This section lists common RISC-V specific toolchain command line options.
+
+### `-mstrict-align`/`-mno-strict-align`
+Control the code generation for scalar memory access with strict alignment (also known as unaligned access or misaligned access).
+
+`-mstrict-align`: The compiler disallows misaligned access for scalar memory access.
```
topperc, line 410: Borrowing from AArch64 gcc documentation. Perhaps

```
indicates that the compiler should not assume that unaligned scalar memory references are handled by the system.
```

README.mkd
```diff
@@ -400,11 +400,22 @@ specification, and in the draft [CORE-V Builtin
 Function](https://github.com/openhwgroup/core-v-sw/blob/master/specifications/corev-builtin-spec.md)
 specification.
 
+## Common Toolchain Command Line Options
+
+This section lists common RISC-V specific toolchain command line options.
+
+### `-mstrict-align`/`-mno-strict-align`
+Control the code generation for scalar memory access with strict alignment (also known as unaligned access or misaligned access).
```
nick-knight, line 408:
```suggestion
Control the code generation for scalar memory access with strict alignment (non-strict also known as unaligned access or misaligned access).
```

nick-knight (approved):
I have no objection to documenting the current behavior.

cmuellner (approved):
LGTM.
Should we also mention that intrinsics are covered, but inline-assembly is not?

kito-cheng:
Update to reflect the discussion within GCC and LLVM community: keep it control both scalar and vector unaligned access, because LLVM already use that semantic for more than one release, change it is not good idea, also GCC 14 isn't really control vector memory by -mstrict-align...it's always allowed in GCC 14, so we'll gonna "fix" GCC 14 to fix this behavior

kito-cheng:
Update:
- Update the option priority for unaligned access.

topperc (approved):
LGTM

kito-cheng:
Gonna merge since I believe we reach consensus between LLVM and GCC community

PR #50, opened 2024-05-29
Add -m[no-][scalar|vector]-strict-align

kito-cheng:
Add two new option `-m[no-]scalar-strict-align` for alias of `-m[no-]-strict-align` and `-m[no-][vector]-strict-align` for vector misaligned access.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -422,6 +422,18 @@ NOTE: The compiler may generate misaligned access if the program violates the
       alignment assumption.
 NOTE: This option does not affect inline assembly.
 
+
+### `-mscalar-strict-align`/`-mno-scalar-strict-align`/`-mvector-strict-align`/`-mno-vector-strict-align`
+
+`-mscalar-strict-align`/`-mno-scalar-strict-align`: Similar to
+`-mstrict-align`/`-mno-strict-align` but applied to scalar memory access only.
+
+`-mvector-strict-align`/`-mno-vector-strict-align`: Similar to
+`-mstrict-align`/`-mno-strict-align` but applied to vector memory access only.
+
+The precedence among `-m[no]-scalar-strict-align`, `-m[no-]vector-strict-align`,
+and `-m[no-]strict-align` is determined by the last one specified.
+
 ## TODO
 
 * `-mdiv`, `-mno-div`, `-mfdiv`, `-mno-fdiv`, `-msave-restore`,


kito-cheng:
Ref:
Discussion in riscv-c-api-doc: https://github.com/riscv-non-isa/riscv-c-api-doc/issues/73
GCC patch: https://patchwork.sourceware.org/project/gcc/patch/cc653588-ac19-4f09-9693-78a4d47cfb1f@gmail.com/

kito-cheng:
CC: @nick-knight @kito-cheng @vineetgarc @asb @aswaterman

topperc:
CC @preames

nick-knight (commented):
I shared my opinion on a related proposal here: https://github.com/riscv-non-isa/riscv-c-api-doc/issues/73#issuecomment-2125480605

I don't personally plan to use these flags, so I'm OK if it remains GCC-specific behavior, but I won't fight it if others in the community think it's actually useful.

README.mkd
```diff
@@ -412,6 +412,15 @@ Control the code generation for scalar memory access with strict alignment (also
 
 NOTE: The compiler may generate misaligned access if the program violates the alignment assumption.
 
+
+### `-mscalar-strict-align`/`-mno-scalar-strict-align`
+
+Aliases for -mstrict-align/-mno-strict-align.
```
cmuellner, line 418: Is this correct? I.e., does `-mstrict-align` not cover vector memory accesses?
If so, then we should add this detail to #49.
If not, then we need to state here, that it only covers scalar memory accesses.

topperc: I believe #49 does say "scalar memory accesses"

cmuellner: Oh, right, it even uses these words multiple times.
There must have been a rendering issue in my browser or so... ;)

cmuellner (approved):
LGTM, but has a dependency for #49.

kito-cheng:
Changes:
- Update the description to match -mstrict-align affect both scalar and vector.

README.mkd
```diff
@@ -422,6 +422,18 @@ NOTE: The compiler may generate misaligned access if the program violates the
...
+### `-mscalar-strict-align`/`-mno-scalar-strict-align`/`-mvector-strict-align`/`-mno-vector-strict-align`
+
+`-mscalar-strict-align`/`-mno-scalar-strict-align`: Similar to
+`-mstrict-align`/`-mno-strict-align` but applied to scalar memory access only.
+
+`-mvector-strict-align`/`-mno-vector-strict-align`: Similar to
+`-mstrict-align`/`-mno-strict-align` but applied to vector memory access only.
+
+The precedence among `-m[no]-scalar-strict-align`, `-m[no-]vector-strict-align`,
+and `-m[no-]strict-align` is determined by the last one specified.t
```
topperc, line 435: Stray 't' at end of line

kito-cheng:
Changes:
- Drop redundant `t` in the line of end.

topperc (approved):
LGTM

PR #52, opened 2024-07-09
Migration to RISC-V specifications template in asciidoc

apazos:
TODOs:
- Fix rendering of Chinese characters in the generated PDF file. Characters referenced in
https://github.com/riscv-non-isa/riscv-toolchain-conventions#list-of-vendor-prefixes

- Fix reference to RISC-V Assembly Programmer's Manual after migration to RVI template is merged.


apazos:
Pushed changes to make title page and toc appear, reorganizing prelude and preamble files

apazos:
Moved list of tables after document status

cmuellner:
I could reproduce the issue with the generation of Chinese characters. @wmat, can you comment on that? What we want to achieve can be seen in the [markdown rendering](https://github.com/riscv-non-isa/riscv-toolchain-conventions/blob/main/README.mkd#list-of-vendor-prefixes) (just search for "Nuclei").

The rest looks good and ready to land.

wmat:
Chinese characters should render fine in the PDF.  In this particular case, if they're only in the README, you can leave it as mkd as it's not included as part of the PDF.

wmat:
The build isn't picking up the theme from docs-resources.  Here's what I get when I build locally:

Screenshot from 2024-08-28 08-57-56

I pushed the fix.

cmuellner:
@wmat, thank you very much! Works now here as well!

cmuellner (approved):
LGTM

apazos:
Thanks a lot, @wmat for fixing the Chinese characters issue.

PR #53, opened 2024-07-29
Add vendor symbols table for vendor-specific relocs

MaelCravero:
See https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/423


kito-cheng (approved):
LGTM :)

PR #54, opened 2024-08-16
Add options for enable code gen with CFI `-fcf-protection=[full|branc…

kito-cheng:
…h|return|none]` and `-mcf-label-scheme=[unlabeled|func-sig]`

Resue the options defined by X86 CET, `-fcf-protection=[full|branch|return|none]`

`-fcf-protection=branch` for landing pad (`Zicfilp`), `-fcf-protection=return` for landing pad (`Zicfiss`) and `-fcf-protection=full` for enable both if possible, landing pad just require instrcution defined by base extension, so compiler will emit landing pad even without `Zicfilp` extension, but `-fcf-protection=return` will require at least `Zimop` since the instrcution isn't included in base extension.

Also we defined another option for specify the labeling scheme: `unlabeled` and `func-sig`.

The `unlabeled` scheme is always use `lpad 0`, and `func-sig` is based on the function signature, the rule is defined in psABI.

---

Currently clang/LLVM using `-fsanitize=shadow-call-stack` to control the shadow stack, also that shared same option with software shadow shadow, so I think we have three options for this:
1. Let `-fsanitize=shadow-call-stack` alias to `-fcf-protection=return`
2. Let  `-fsanitize=shadow-call-stack` only for software shadow stack
3. Drop `-fcf-protection=return` and only `-fsanitize=shadow-call-stack`

I am OK with option 1 or 2 and prefer option 2, and dislike option 3 since it means we have very different flavor option naming scheme on landing pad and shadows stack...


mylai-mtk:
I believe all mentions of the `simple` scheme would be changed to `unlabeled` in the future?

README.mkd
```diff
@@ -445,6 +445,28 @@ NOTE: This option does not affect inline assembly.
...
+
+- `none`: Disable control flow protection.
+- `full`: Protect all control flow instructions, will enable branch protection
+  and return protection if the `Zimop` extension is available.
+- `branch`: Protect branch instructions only by insert landing pad.
+- `return`: Protect branch instructions only, this require `Zimop` extension.
+
+### `-mcf-label-scheme=[simple|func-sig]`
+
+Specify the label scheme for the `-fcf-protectio=branch`. The default is value
```
mylai-mtk, line 463: Typo: -fcf-protectio'n'=branch

kito-cheng:
> I believe all mentions of the simple scheme would be changed to unlabeled in the future?

Oh, yeah, apparently I lost my memory when I wrote this PR

[1] https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/417#issuecomment-2232684743

jaidTw:
Do we have conclusion on which option to implement now?

mylai-mtk:
> Do we have conclusion on which option to implement now?

I'm good with `-fcf-protection=branch|full` and `-fcf-branch-label-scheme=unlabeled|func-sig`. I tried implementing it in Clang, and currently there's no problem.

Besides the adoption of this `-fcf-protection=branch` flag, I want to ask if we're going implement the `[no]cf_check` attribute as introduced in [x86 Function Attributes](https://gcc.gnu.org/onlinedocs/gcc/x86-Function-Attributes.html#index-nocf_005fcheck-function-attribute)?

src/toolchain-conventions.adoc
```diff
@@ -406,6 +406,29 @@ NOTE: This option does not affect inline assembly.
...
+
+Enable control flow protection. The compiler will insert control flow integrity
+instructions to protect the program against control flow hijacking attacks.
+
+`-fcf-protection` is alias to `-fcf-protection=full`.
+
+- `none`: Disable control flow protection.
+- `full`: Protect all control flow instructions, will enable branch protection
+  and return protection if the `Zimop` extension is available.
+- `branch`: Protect branch instructions only by insert landing pad.
```
topperc, line 420: insert -> inserting

src/toolchain-conventions.adoc
```diff
@@ -406,6 +406,29 @@ NOTE: This option does not affect inline assembly.
...
+=== `-fcf-protection=[full|branch|return|none]`/`-fcf-protection`
+
+
+Enable control flow protection. The compiler will insert control flow integrity
+instructions to protect the program against control flow hijacking attacks.
+
+`-fcf-protection` is alias to `-fcf-protection=full`.
+
+- `none`: Disable control flow protection.
+- `full`: Protect all control flow instructions, will enable branch protection
```
topperc, line 418: This sentence reads like branch protection requires Zimop, but I think only return protection requires Zimop?

quic-garvgupt:
Hi @kito-cheng,

To clarify our last discussion during the RISCV sync-up, does `-fcf-protection=return` now imply `-fsanitize=shadow-call-stack`? Additionally, according to the documentation, `-fcf-protection=return` requires the Zimop extension to be enabled. If this extension is not enabled, will a software shadow stack be generated instead?

kito-cheng:
@quic-garvgupt

At LLVM sync up meeting, we would like use `-fsanitize=shadow-call-shadow` for software shadow stack, and `-fcf-protection=return` for hardware shadow stack only, one option control one feature, so no worry about zimop disturb `-fsanitize=shadow-call-shadow`  :)

quic-garvgupt:
Thank you, @kito-cheng, for your response. I have one more clarification: when referring to `-fsanitize=shadow-call-shadow`, do we actually mean `-fsanitize=shadow-call-stack`? I was unable to find any option named `-fsanitize=shadow-call-shadow`, and Clang returns an “unsupported option” error. Is this a new option that will be supported in the future, or is it simply a typographical error?

kito-cheng:
Both GCC[1] and LLVM[2] has landed, so I believe this PR is ready to land :) @cmuellner could you help me to merge this? 

[1] https://github.com/gcc-mirror/gcc/commit/59a869d7196132ae5257fcb068508561d3526c7d
[2] https://github.com/llvm/llvm-project/pull/109784

cmuellner (approved):

PR #55, opened 2024-08-28
Create build-pdf.yml

wmat:
Add an action to build the pdf spec.


cmuellner (approved):

cmuellner:
Thanks for the PR!

I think we would prefer to have the workflow more like the riscv-elf-psabi-doc repo. However, let's start with something and adjust as we go.

wmat:
I'd recommend sticking with the Docker based build as that way you'll be sure to be using the latest document toolchain.

PR #56, opened 2024-09-03
Add SiFive VCIX state register

4vtomat:


src/toolchain-conventions.adoc
```diff
@@ -367,6 +367,14 @@ specification, and in the draft
 https://github.com/openhwgroup/core-v-sw/blob/master/specifications/corev-builtin-spec.md[CORE-V Builtin Function]
 specification.
 
+=== List of vendor CSRs
+.List of vendor CSRs
+[%autowidth]
+|===
+|*Vendor*              |*Name*             |*Description*
+|SiFive                |sf.vcix_state     |This dummy CSR is used to prevent the scheduler to reorder between SiFive VCIX instructions.
```
kito-cheng, line 375: Maybe mention XSFVCP?
```suggestion
|SiFive                |sf.vcix_state     |This dummy CSR is used to prevent the scheduler to reorder between SiFive vector instructions defined in `XSFVCP` extension.
```

src/toolchain-conventions.adoc
```diff
@@ -367,6 +367,14 @@ specification, and in the draft
...
 specification.
 
+=== List of vendor CSRs
+.List of vendor CSRs
+[%autowidth]
+|===
+|*Vendor*              |*Name*             |*Description*
+|SiFive                |sf.vcix_state     |This dummy CSR is used to prevent the scheduler to reorder between SiFive vector instructions defined in `XSFVCP` extension.
+|===
+
```
cmuellner, line 377: This document should not become a registry of all vendor CSRs (or vendor instructions).
Instead, let's restrict ourselves to vendor prefixes and vendor extensions with the intent to avoid potential name collisions.

We've defined the vendor prefixes to be equal for instructions and CSRs (see the section "CSR naming scheme" above). A SiFive vendor prefix (`sf.`) and the `XSFVCP` extension are already defined above. Therefore, I think this PR is not needed.

kito-cheng:
That's kinda special case for the vendor CSR, it's a pseudo register which used for guarantee the order between those instruction, but I agree that should not list here, otherwise other vendor will go here to add their own, and this doc will become hard to maintain.

PR #59, opened 2024-09-09
Add large code model

tclin914:


cmuellner:
This looks good in general.

When double-checking with the implementations, I saw that LLVM does not create `__riscv_cmodel_large`. We should probably close this gap before merging this PR.

cmuellner (approved):
LGTM

cmuellner:
@asb and @kito-cheng: any objections to merging this?

It looks like this is needed for https://github.com/llvm/llvm-project/pull/107817

kito-cheng (approved):
LGTM :)

asb:
In parallel I'd added the define to riscv-c-api-doc in https://github.com/riscv-non-isa/riscv-c-api-doc/pull/86

This listing of defines comes from when this repo was little more than my personal notes on what GCC does and Clang should probably do to follow suit. Now that we have the preprocessor defines listed authoritatively in riscv-c-api-doc I think it makes sense to remove them from this document, which https://github.com/riscv-non-isa/riscv-toolchain-conventions/pull/61 does.

PR #63, opened 2024-11-19
Added Qualcomm Microcontroller Extensions

apazos:
We have added 15 extensions to the Microcontroller Integer Unit.

The specifications are publicly available at
https://github.com/quic/riscv-unified-db/releases

The development of the ELF psABI with Qualcomm extensions is currently underway.

Support for these extensions in the LLVM toolchain is also in progress.


cmuellner:
Thanks for the PR!

The referenced extensions are listed as version 0.1. However, when looking at the referenced repo, the individual extensions don't have a version. For the XThead* extensions we simply set v1.0 for all extensions, so it is clear they won't ever change.

The referenced document has a "DRAFT" watermark, which usually implies that the extensions are not frozen. If that's the case, then this is a blocker for upstreaming.

apazos:
Thanks, Christopher, for the quick feedback.

Although the spec is labeled as a draft, it is currently in a stable state. The upcoming PDF release will reflect this status, and the list of extensions and their instructions is also considered stable.

We have assigned version 0.1 to the extensions because we are in the process of implementing assembler support for the instructions in LLVM. This allows us to correct any typos and minor issues in the spec if needed. Once this process is complete, the extensions and spec will be promoted to version 1.0.

As we begin to upstream support in LLVM under the experimental extension, we have decided to introduce the Qualcomm vendor extensions in this spec. The patches will refer to Qualcomm extension names, use the "qc" prefix for instructions, and include Qualcomm vendor relocations. This is the spec that should contain such information.

apazos:
@cmuellner, the new frozen release is out: https://github.com/quic/riscv-unified-db/releases.
We will promote it to 1.0 once we finalize the assembler support in the LLVM toolchain.

apazos:
Will push 2 separate patches, one for the vendor prefix and the other for the extensions

PR #71, opened 2025-01-14
Propose vendor mnemonic extension for CHERIoT

resistor:
Cheriot (https://cheriot.org/) is a RISC-V-based microcontroller platform that incorporates CHERI technology for capability-based security. It extends RISC-V with several new instructions (found in the ISA reference on cheriot.org). Some of these share names with the Zcheri extension currently under consideration, but do not use the same encodings, and thus should be in a vendor prefix.

The Cheriot LLVM toolchain is maintained in https://github.com/cherIoT-Platform/llvm-project It does not currently use a prefix on its custom instructions, but we would like to change that if we can assign a prefix.

### DIFF
--- a/src/toolchain-conventions.adoc
+++ b/src/toolchain-conventions.adoc
@@ -277,6 +277,7 @@ separated by a dot, e.g., `th.vxrm`.
 [%autowidth]
 |===
 |*Vendor*              |*Prefix*         |*URL*
+|Cheriot                | ct              | https://cheriot.org/
 |Open Hardware Group    | cv              | https://www.openhwgroup.org/
 |Andes                  | nds             | https://www.andestech.com/
 |SiFive                 | sf              | https://www.sifive.com/


apazos:
Is there a plan to upstream these additional custom instructions to LLVM?
I understand the reservation of vendor acronyms and instructions/CSRs prefixes is to avoid conflicts in upstream LLVM and Clang.

resistor:
We (SCI Ltd, who is shipping Cheriot HW in 2025 and driving Cheriot development) intend to propose upstreaming Cheriot support to LLVM/Clang this year.

We're currently actively trying to catch up our repository to top of tree, and we'd like to start getting our RISC-V extension into compliance with these toolchain conventions to facilitate upstreaming.

apazos:
Thanks for the clarification @resistor. 
The request sounds reasonable. 
@cmuellner, I will go ahead and merge it. Let me know if you see any problem.

apazos:
We should also reorganize the list in alphabetic order. Might be cleaner to do in separate commit from yours, @resistor.

apazos:
Also, @resistor, do you have the link to the spec where the additional custom instructions are defined?

apazos:
I found some spec at https://cheriot.org/cheriot-sail/cheriot-architecture.pdf as version 0.6 drat, is it in frozen state?

resistor:
> I found some spec at https://cheriot.org/cheriot-sail/cheriot-architecture.pdf as version 0.6 drat, is it in frozen state?

That is the live draft which is continuously updated from the SAIL model here: https://github.com/CHERIoT-Platform/cheriot-sail
We expect to freeze v1 in the next few months.

resistor:
> We should also reorganize the list in alphabetic order. Might be cleaner to do in separate commit from yours, @resistor.

It appears to already be alphabetical by prefix. I was just following that convention.

cmuellner (approved):

apazos (approved):

PR #72, opened 2025-01-24
Introduce vendor prefix for Rivos custom extensions

preames:
This change reserves the vendor prefix "ri." for Rivos's custom extensions.

### DIFF
--- a/src/toolchain-conventions.adoc
+++ b/src/toolchain-conventions.adoc
@@ -280,6 +280,7 @@ separated by a dot, e.g., `th.vxrm`.
 |Cheriot                | ct              | https://cheriot.org/
 |Open Hardware Group    | cv              | https://www.openhwgroup.org/
 |Andes                  | nds             | https://www.andestech.com/
+|Rivos                  | rv              | https://www.rivosinc.com/
 |SiFive                 | sf              | https://www.sifive.com/
 |T-Head                 | th              | https://www.t-head.cn/
 |Tenstorrent            | tt              | https://www.tenstorrent.com/
--- a/src/toolchain-conventions.adoc
+++ b/src/toolchain-conventions.adoc
@@ -280,7 +280,7 @@ separated by a dot, e.g., `th.vxrm`.
 |Cheriot                | ct              | https://cheriot.org/
 |Open Hardware Group    | cv              | https://www.openhwgroup.org/
 |Andes                  | nds             | https://www.andestech.com/
-|Rivos                  | rv              | https://www.rivosinc.com/
+|Rivos                  | ri              | https://www.rivosinc.com/
 |SiFive                 | sf              | https://www.sifive.com/
 |T-Head                 | th              | https://www.t-head.cn/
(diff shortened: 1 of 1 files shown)


TommyMurphyTM1234:
Might "rv" perhaps get confused with official "RISC-V" standard extensions?

preames:
> Might "rv" perhaps get confused with official "RISC-V" standard extensions?

The mnemonics for standard extensions are not prefixed.

TommyMurphyTM1234:
> > Might "rv" perhaps get confused with official "RISC-V" standard extensions?
> 
> The mnemonics for standard extensions are not prefixed.

I appreciate that, but isn't it possible that some casual observers won't know that and might mistakenly assume that "rv" prefixed extensions might be official/standard/non-proprietary RISC-V? Especially since "RV" is so widely used as an informal way to refer to RISC-V features in general. An easy way to eliminate the possibility of such confusion might be by using a different prefix here?

cmuellner:
We talked about this in the SIG Toolchain call today.

I acknowledge the potential for confusion with an `rv.` prefix.
However, vendor code is not expected to be emitted without explicitly enabling it (similar arguments apply to other occurrences of vendor instructions like hand-written assembly).

Where it might cause confusion is in the intrinsic names.
There, we have a prefix of `__riscv_` followed by the vendor prefix.
With accepting `rv`, we could get intrinsics like `__riscv_rv_vadd_vv_i8m1()` (similar to `__riscv_vadd_vv_i8m1`).

I stated that I wouldn't block this PR, but I'd like to hear opinions from GCC and LLVM maintainers on this.

kito-cheng:
> I acknowledge the potential for confusion with an `rv.` prefix. However, vendor code is not expected to be emitted without explicitly enabling it (similar arguments apply to other occurrences of vendor instructions like hand-written assembly).
> 
> Where it might cause confusion is in the intrinsic names. There, we have a prefix of `__riscv_` followed by the vendor prefix. With accepting `rv`, we could get intrinsics like `__riscv_rv_vadd_vv_i8m1()` (similar to `__riscv_vadd_vv_i8m1`).
> 
> I stated that I wouldn't block this PR, but I'd like to hear opinions from GCC and LLVM maintainers on this.

I was also hesitate about that may confused by riscv as well, but that sound reasonable with those two reasons.

kito-cheng:
So yeah, LGTM :)

preames:
FYI, triggered by the discussion here and in the LLVM reviews, we have changed the specification for the instructions (https://github.com/rivosinc/rivos-custom-extensions) to use the "ri." prefix instead of "rv."  To date, only experimental support has landed in LLVM and we just changed that over as well.  

@kito-cheng If you're good with the change, could you land the new version?

cmuellner (approved):

cmuellner:
Related LLVM PR: https://github.com/llvm/llvm-project/pull/128761

cmuellner:
LLVM merged the `ri` vendor prefix for Rivos.
So, let's land this as well.

PR #73, opened 2025-02-10
Vendor extensions: Add SpacemiT vendor prefix

zqb-all:
This patch defines the SpacemiT instruction prefix `smt`.


kito-cheng (commented):
LGTM

asb:
Nit: Patch title should be something like "Vendor extensions: Add SpacemiT vendor prefix" as the patch just defines the prefix and doesn't add any extensions.

zqb-all:
> @zqb-all do you have a link to the spec that describes the extensions you plan to add in the future?

The [SpacemiT IME extension](https://github.com/space-mit/riscv-ime-extension-spec) is already supported in the SpacemiT k1 chip. However, the instructions in this document don't include vendor prefix yet, so I'm thinking of defining a prefix first and then adding a PR after the vendor prefix is added to the documentation

cmuellner (approved):
LGTM

cmuellner:
Let's land this.

PR #74, opened 2025-02-18
Add more SiFive extensions to List of vendor extensions

topperc:
Two extensions start with XSiFive for historical documentation reasons. Newer extensions use XSF.


src/toolchain-conventions.adoc
```diff
@@ -311,34 +311,37 @@ tag to provide extra relocations for a given vendor.
...
+|OpenHW    | Xcvelw               | 1.0.0    | https://github.com/openhwgroup/cv32e40p/blob/dev/docs/source/instruction_set_extensions.rst[CORE-V Instruction Set Extensions]
+|OpenHW    | Xcvhwlp              | 1.0.0    | https://github.com/openhwgroup/cv32e40p/blob/dev/docs/source/instruction_set_extensions.rst[CORE-V Instruction Set Extensions]
+|OpenHW    | Xcvmac               | 1.0.0    | https://github.com/openhwgroup/cv32e40p/blob/dev/docs/source/instruction_set_extensions.rst[CORE-V Instruction Set Extensions]
+|OpenHW    | Xcvmem               | 1.0.0    | https://github.com/openhwgroup/cv32e40p/blob/dev/docs/source/instruction_set_extensions.rst[CORE-V Instruction Set Extensions]
+|OpenHW    | Xcvsimd              | 1.0.0    | https://github.com/openhwgroup/cv32e40p/blob/dev/docs/source/instruction_set_extensions.rst[CORE-V Instruction Set Extensions]
+|SiFive    | XSiFivecdiscarddlone | 1.0      | https://sifive.cdn.prismic.io/sifive/767804da-53b2-4893-97d5-b7c030ae0a94_s76mc_core_complex_manual_21G3.pdf[SiFive S76-MC Core Complex Manual]
```
cmuellner, line 323: If I open the referenced document, I can't find the term "XSiFivecdiscarddlone" in there.

TommyMurphyTM1234: > If I open the referenced document, I can't find the term "XSiFivecdiscarddlone" in there.

Isn't this it? 

> **3.4.2 Cache Maintenance Operations**
>
> The data cache supports `CFLUSH.D.L1` and `CDISCARD.D.L1`. The instruction `CFLUSH.D.L1`
cleans and invalidates the specified line or all cache lines. The instruction `CDISCARD.D.L1` invalidates the specified line or all cache lines.
> These custom instructions are further described in Chapter 6.

topperc: Yes, those are the instructions. The extension name didn't make it into that document, but it's name has been used in SiFive's internal toolchain for several years.

cmuellner: I don't think this table should define the name and version of vendor extensions. The idea of the table is to have an index.

My preferred solution would be for SiFive to host a document where the extension is defined (name, version, instructions, CSRs). That's the bar we also had for other extensions in the table.

PR #77, opened 2025-02-28
Added Qualcomm Microcontroller extensions (Xqci, Xqccmp)

apazos:
We have added 18 extensions to the Microcontroller Integer Unit, collectively referred to as Xqci extensions.

In addition we have added the Xqccmp extension with push/pop and double move instructions.

The specifications are publicly available at
https://github.com/quic/riscv-unified-db/releases

The psABI with Qualcomm extensions is available at https://github.com/quic/riscv-elf-psabi-quic-extensions

Support for these extensions in the LLVM toolchain is also in progress.


apazos:
@cmuellner, this is the second commit which includes the list of microcontroller extensions whose spec is in frozen state. The PDF document still has some rendering issues, once resolved it will be promoted to 1.0.

apazos:
@cmuellner, we can go ahead and merge this change. Implementation is progressing in LLVM and Qualcomm's open source ELD linker is about to commence.

kito-cheng (approved):

PR #78, opened 2025-03-11
Removed version info from the table of vendor extensions.

apazos:
The version number is not relevant for this table. However, to include an extension to this table, the ISA documentation must reference a stable or frozen version of the extension that is ready for implementation in the open-source toolchains like LLVM and GCC.


cmuellner (approved):

cmuellner:
LGTM, but there is a merge conflict that needs to be resolved.

PR #80, opened 2025-03-31
Added policy for adding vendor prefixes and vendor extensions

apazos:


apazos:
Based on the feedback collected during the Toolchains SIG meeting held on 03/27/2025, a policy has been added regarding the allocation of vendor prefixes and the listing of vendor extensions within this document.

kito-cheng (approved):
LGTM as a RISC-V GCC maintainer

cmuellner:
The policy looks fine, but I would prefer to have it in a different file (it is not part of the toolchain conventions but a policy about changes to that document).

lenary (approved):
LGTM. 

I recall we mentioned that there was another place that also stated policy (not contents) in this document. I think we can move those out (or into an appendix?) in a separate PR?

cmuellner:
> I recall we mentioned that there was another place that also stated policy (not contents) in this document. I think we can move those out (or into an appendix?) in a separate PR?

I mentioned the psABI repo, where we store the merge policy in [policy.md](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/policy.md).

apazos:
It is possible to relocate the policy text to a separate file; however, it may be beneficial to keep it close to the prefix/extensions tables for visibility, or a reference to the policies must be maintained close to the tables. 
Also note that the vendor extensions table will be moved to the appendix after merging the policies.

apazos:
But the policy file will be part of the toolchain conventions document.

cmuellner:
> But the policy file will be part of the toolchain conventions document.

Having it in the repo is the right thing. Separating from the document (but keeping it in the repo) is motivated by the fact that it is about changing the document, but not the document itself.

apazos:
We discussed at the last meeting how things are already interwoven in the document. But putting the policies it a separate file in this repo makes sense to me. Will push the change.

apazos:
@cmuellner, I added a separate policy file. 
I noted that the references to the file are failing in the preview. I think this issue should be resolved after the file is merged into the repo. It works correctly locally and in my test branch https://github.com/apazos/riscv-toolchain-conventions/tree/mytest.

Please review and merge. If there is any remaining reference issue, I fix it in a follow up patch.

In a future patch, I will move the vendor extensions list to the appendix, as discussed.

cmuellner (approved):
LGTM, thank you!

apazos:
@cmuellner, Please go ahead and merge.

apazos:
@cmuellner, I went ahead and replaced the xref with the URL link to the policy, it should work from the pdf. I will be able to test once the policy file is merged.

cmuellner (approved):
Like it very much!
Thanks!

PR #83, opened 2025-04-16
Alphabetized the prefix and extensions lists by vendor name

apazos:


cmuellner:
Thanks for catching this!

There are two tables:
* vendor, prefix, link -> sorted by prefix
* vendor, extension, link -> sorted by vendor

The sort order seems arbitrary.
It would be better to sort both by vendor or by prefix/extension.

Also, since the focus is on the prefix/extension, we could swap the first two columns of the table.

apazos:
I think that ordering vendor extensions by name makes the table less readable due to its length. 
Might be best to just order both by vendor.

cmuellner:
> I think that ordering vendor extensions by name makes the table less readable due to its length. Might be best to just order both by vendor.

I can see pros/cons for both orderings.
If needed, a Ctrl + F will help to find entries.

Thank you for ordering both tables the same way!

PR #86, opened 2025-05-08
Moved list of vendor extensions to the appendix

apazos:
The list of vendor extensions provides an informational reference for the extensions supported by vendors. We discussed moving it to an appendix at the Toolchains SIG meeting held on 03/27/2025.


cmuellner (approved):
LGTM

apazos:
thanks Christoph!

PR #91, opened 2025-07-03
Add Target Triple Architecture Names

lenary:
As discussed at today's toolchains SIG.

### DIFF
--- a/src/toolchain-conventions.adoc
+++ b/src/toolchain-conventions.adoc
@@ -1,3 +1,17 @@
+== Target Triples
+
+The toolchains have agreed on the following architecture names to be used for
+RISC-V in the first segment of target triples:
+
+- `riscv32` RV32, little-endian
+- `riscv32be` RV32, big-endian
+- `riscv64` RV64, little-endian
+- `riscv64be` RV64, big-endian
+
+For example, `riscv32-unknown-none-elf` denotes 32-bit RISC-V, little-endian,
+from an unknown vendor, for no Operating System (baremetal), following the ELF
+ABI.
+
 == Specifying the target ISA with -march
 
 The compiler and assembler both accept the `-march` flag to specify the target


src/toolchain-conventions.adoc
```diff
@@ -1,3 +1,17 @@
+== Target Triples
```
cmuellner, line 1: The Autoconf manual speak of "Triplets" not "Triples" (see https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Specifying-Target-Triplets.html).

src/toolchain-conventions.adoc
```diff
@@ -1,3 +1,17 @@
...
+
+The toolchains have agreed on the following architecture names to be used for
+RISC-V in the first segment of target triples:
+
+- `riscv32` RV32, little-endian
+- `riscv32be` RV32, big-endian
+- `riscv64` RV64, little-endian
+- `riscv64be` RV64, big-endian
+
```
cmuellner, line 10: config.sub in the autoconf repo also knows `riscv` (see https://github.com/autotools-mirror/autoconf/blob/master/build-aux/config.sub#L1239).
But I don't know if that is supported by tools.

src/toolchain-conventions.adoc
```diff
@@ -1,3 +1,17 @@
+== Target Triples
+
+The toolchains have agreed on the following architecture names to be used for
```
cmuellner, line 3: I would prefer a different wording.
E.g. "The following architecture names are used for RISC-V in the first segment of target triplets".

src/toolchain-conventions.adoc
lenary: Clang doesn't seem to know about it:
```
$ clang --target=riscv -march=rv32gc -print-effective-triple
riscv
$ clang --target=riscv64 -march=rv32gc -print-effective-triple
riscv32--
```

(What this shows is that clang is canonicalising the triplet when the march/triplet arch disagree with the 63/32 suffix, but isn't touching the triplet arch at all when the arch is just riscv, because it is not recognised). The llvm Triple(t) docs point to `config.guess`, which doesn't seem to mention just `riscv`. The code to make the triplet switch based on `-march` is here: https://github.com/llvm/llvm-project/blob/2a702cdc38f96989cb41c0b863d977f0d8f4f325/clang/lib/Driver/Driver.cpp#L845

src/toolchain-conventions.adoc
lenary: Sure, that sounds better.

src/toolchain-conventions.adoc
lenary: Will do.

lenary:
I found this about `riscv32e`: https://reviews.llvm.org/D53291 (never landed as far as I can tell).

Rust uses triple-like things in its target names, and has a [`riscv32e` target](https://doc.rust-lang.org/beta/rustc/platform-support/riscv32e-unknown-none-elf.html) but note that in this case the target name includes `-march`/`-mabi` information which C/C++ toolchains tend not to

cmuellner:
@kito-cheng and @topperc, any objections?

PR #98, opened 2025-07-31
Document -march=unset option behavior and usage

kito-cheng:
Add documentation for the -march=unset option that allows ignoring previous -march options and deriving architecture from -mcpu.

The -march=unset option is particularly useful for build systems and toolchain configurations where architecture should always be derived from CPU specification rather than potentially conflicting -march options.


lenary (approved):
LGTM. Thanks Kito!

kito-cheng:
GCC patch has landed, so I believe it's settle down now :)

MaskRay:
GCC commit 4a182418c89666e7594bcb0e5edc5194aa147910 ("RISC-V: Support CPUs in -march.") has not been reverted yet but should be reverted.

PR #134, opened 2025-12-16
Add vendor relocation identifier for Cheriot.

resistor:


src/toolchain-conventions.adoc
```diff
@@ -425,6 +425,7 @@ tag to provide extra relocations for a given vendor.
 |===
 |*Vendor*              |*Symbol*
 |Andes                 | ANDES
+|Cheriot               | XCHERIOT
```
jrtc27: The X is a bit superfluous here given this is specifically for vendor relocations, and the X in extension names is just a prefix to state that what follows is a vendor extension.

jrtc27: (See how none of the others include it)

resistor: Done

resistor:
I think someone else has to merge this on my behalf.

cmuellner (approved):
LGTM

PR #139, opened 2026-01-14
Add -mcpu Extension Control Syntax for RISC-V Toolchain

majin2020:
## Implementation Overview

This patch adds comprehensive `-mcpu` extension control syntax support to the RISC-V toolchain specification, following ARM GCC's mature design pattern.

## Design References

- **ARM Clang Documentation**: [https://developer.arm.com/documentation/107976/21-1-1/Clang-reference/clang-command-line-options/-mcpu](https://developer.arm.com/documentation/107976/21-1-1/Clang-reference/clang-command-line-options/-mcpu)
- **ARM GCC Documentation**: [https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html#index-mcpu](https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html#index-mcpu)

## Background and Motivation

In real-world RISC-V ecosystems, the same CPU IP core often exists in multiple hardware configuration variants:

- **Floating-point units**: With F/D extensions vs without floating-point units
- **Vector units**: With V extensions vs without vector units
- **Vector length**: Configurable VLEN (e.g., 128b, 256b, 512b, etc.)
- **Other extensions**: Optional components like Zicsr, Zifencei, compressed instructions, etc.

If we were to define separate CPU model names for every configuration combination, it would lead to exponential growth in CPU model names, creating a significant maintenance burden for toolchains and usability challenges for developers.

### 1. Extension Control Syntax
- Supports `+extension` / `+noextension` syntax to enable/disable specific RISC-V extensions
- Rightmost precedence rule: conflicting modifiers are resolved right-to-left
- Supports all standard and vendor-specific RISC-V extensions

### 2. Dependency Handling
- Properly handles extension dependencies (e.g., D extension depends on F extension)
- Automatically disables dependent extensions when base extensions are disabled

## Problems Solved

1. **Model explosion problem**: Avoids creating numerous duplicate CPU model definitions for different configurations of the same CPU IP
2. **Configuration flexibility**: Users can precisely control processor extensions without being limited to predefined CPU models
3. **Hardware compatibility**: Enables compilation for processor variants missing certain optional extensions

## Usage Examples

```bash
# Disable floating-point extensions (for xt-c908 variants without FPU)
gcc -mcpu=xt-c908+nof

# Enable vector extensions (equivalent to xt-c908v, for variants with V unit)
gcc -mcpu=xt-c908+v+zvfh

# Conflict resolution (right-to-left precedence)
gcc -mcpu=xt-c908+nof+f  # F extension is ultimately enabled

# Multiple extension combinations (for specific VLEN configurations)
gcc -mcpu=xt-c908+v+zvl256b+zvfh
```

This implementation ensures full compatibility with existing GCC RISC-V backends while providing clear and accurate documentation guidance, effectively addressing the CPU model fragmentation issue in the RISC-V ecosystem.

### DIFF
--- a/src/toolchain-conventions.adoc
+++ b/src/toolchain-conventions.adoc
@@ -140,6 +140,56 @@ Examples:
 
 `-march=rva22u64` equals `-march=rv64gcb_zic64b_zicbom_zicbop_zicboz_ziccamoa_ziccif_zicclsm_ziccrse_zicntr_zihpm_za64rs_zfhmin_zkt`
 
+== Specifying the target CPU with -mcpu
+
+The `-mcpu=<name>[+<extension>]*` option uses the architecture of and optimizes the output for the given RISC-V processor.
+
+- The target RISC-V architecture (as if specified by `-march`)
+- The RISC-V processor type for which to tune for performance (as if specified by `-mtune`)
+
+Note that `-mcpu` does not override `-march` or `-mtune`.
+
+RISC-V processors may support optional architectural extensions beyond their base configuration. The `-mcpu` option uses the base configuration defined for the specified CPU, and additional extensions can be enabled using the `+extension` syntax.
+
+=== Extension control syntax
+
+The extension control syntax uses a `+extension` format to enable optional RISC-V extensions:
+
+- `+extension`: Enables the extension if it is not already enabled by default.
+
+Use `+extension` to explicitly enable an optional RISC-V extension.
+You can specify one or more extensions that a processor supports.
+
+Common extension control options include:
+
+- `+f`: Enables the F extension (single-precision floating-point)
+- `+d`: Enables the D extension (double-precision floating-point)
+- `+v`: Enables the V extension (vector instructions)
+- `+zvl256b`: Enables the Zvl256b extension (256-bit vector length)
+- `+zvfh`: Enables the Zvfh extension (half-precision vector floating-point)
+
+The `-mcpu` option supports enabling any valid RISC-V extension using the `+extension` syntax. The examples above illustrate commonly used extensions, but the mechanism applies to all standard and vendor-specific RISC-V extensions.
+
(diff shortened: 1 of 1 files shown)


majin2020:
> I personally don't have any problems with a purely additive extension mechanism.
> 
> I have found that vendors prefer to have `-mcpu=core` denote their "common" configuration for that core and not the "minimal" configuration, so I'm not sure it would be acceptable to others. The driving force behind that is benchmark numbers are (usually) better if one generates code for all the features in their core, and vendors want the best numbers with `-mcpu=<core>` because that's what most of their users will use, rather than the specific `-mcpu` for their exact core configuration.

Hi，thanks for your feedback and suggestions!

I have completely removed the `+noextension` extension-disabling mechanism. The `-mcpu` extension control syntax is now purely additive, supporting only `+extension` to enable additional optional extensions.

Key changes made:
1. **Removed all extension-disabling capabilities**: No longer supports `+nof`, `+nod`, or similar syntax
2. **Simplified the design**: Avoids the complex conflict handling issues seen in the ARM model
3. **Clarified the use case**: `-mcpu=<core>` uses the base configuration defined for the CPU, while `+extension` allows users to enable optional extensions that the CPU supports but are not included in the base configuration

Regarding your concern about vendor preferences, based on my understanding, what we typically need is a "minimal" baseline configuration for `-mcpu=<core>`, with optional features (like vlen256, vlen512, matrix, etc.) layered on top as needed. If other vendors have requirements for a "common" configuration approach, they can certainly bring it up for further discussion. For now, I believe the "minimal" configuration is sufficient.

This approach maintains backward compatibility while providing flexibility for scenarios requiring precise extension control, without the complexity of disabling extensions.

Do you have any other comments or suggestions on this purely additive design? Looking forward to your feedback!

PR #140, opened 2026-01-27
Document Existing Linker Emulation Strings

lenary:
Linkers usually use emulation strings rather than target triples. This documents the existing emula

I have included:
- The little endian emulation strings: `elf32lriscv`/`elf64lriscv` (supported by GNU and Clang/LLD)
- The big endian emulation strings: `elf32briscv`/`elf64briscv` (supported by GNU only at this time)

I have not documented any emulation strings with ABI suffixes, as used by GNU, as Clang/LLD has taken a specific decision not to support these: https://reviews.llvm.org/D95755

If we decide more emulation strings are needed, they can come in a follow-up change.

### DIFF
--- a/src/toolchain-conventions.adoc
+++ b/src/toolchain-conventions.adoc
@@ -12,6 +12,19 @@ For example, `riscv32-unknown-none-elf` denotes 32-bit RISC-V, little-endian,
 from an unknown vendor, for no operating system (baremetal), following the ELF
 ABI.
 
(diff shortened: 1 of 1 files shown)


apazos:
@lenary, are we introducing GNU-specific elements into the Toolchain Conventions by listing these emulation types? If linkers aren’t required to support them, what is the value in documenting them here?

PR #145, opened 2026-02-12
Document --print-supported-extensions and --print-enabled-extensions

lenary:
This documentation is based on the implementation in Clang.


lenary:
IIRC, the `-mcpu=help` lists take into account 32- vs 64-bit CPUs (depending on the triple given to clang). I don't think that `--print-supported-extensions` does so, it prints everything (this difference is less relevant to `--print-enabled-extensions`).

lenary:
Thanks for pulling that out. Yes, this is sort of a suggestion that we could adopt this as a convention, which would require implementing this in gcc (or another tool/script in the gnu toolchain).

PR #156, opened 2026-03-13
Zilsd Alignment Control Flags

lenary:
This codifies the PR/proposal currently open for clang: https://github.com/llvm/llvm-project/pull/181439

### DIFF
--- a/src/toolchain-conventions.adoc
+++ b/src/toolchain-conventions.adoc
@@ -486,6 +486,22 @@ NOTE: This option does not affect inline assembly.
 The precedence among `-m[no]-scalar-strict-align`, `-m[no-]vector-strict-align`,
 and `-m[no-]strict-align` is determined by the last one specified.
 
+=== `-mzilsd-word-align`/`-mzilsd-strict-align`
+
+Zilsd is an extension for making 64-bit accesses on RV32. As with other
+accesses, the spec only guarantees this will work for naturally-aligned accesses
+(8-byte alignment). Some RV32 cores may have a memory system that relaxes this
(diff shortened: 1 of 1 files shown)


kito-cheng (approved):
LGTM :)

kito-cheng (commented):
Few comment around the naming, not blocking comment, Zilsd in theory can be extend to RV64 in future, and `-mzilsd-word-align` might be awkwardness when we have that, but that might not happen soon, so I am fine with that.

lenary:
> Few comment around the naming, not blocking comment, Zilsd in theory can be extend to RV64 in future, and `-mzilsd-word-align` might be awkwardness when we have that, but that might not happen soon, so I am fine with that.

I really hope they don't call the rv64 version "zilsd", given that `LD` and `SD` exist on RV64 already, and iirc there's encoding stuff where they overlap.

The implementation in clang explicitly rejects the options on rv64, so we can always accept other spellings there in the future without breaking users.

christian-herber-nxp:
an RV64 version of this extension is very unlikely. When starting Zilsd, RV64 was quite explicitly excluded, as it doesn't work well to out of order machines with register renaming. Also, macro op fusion exists for such machines, and code density is less of a concern.
If such an extension was made for RV64, it would have to be called Zilsq. However, lq and sq encodings were never allocated for RV128, and I doubt they ever will be. This makes it more unlikely.

cmuellner:
GCC patch on list: https://gcc.gnu.org/pipermail/gcc-patches/2026-May/717222.html

PR #183, opened 2026-09-14
Propose save/restore routines for use with shadow stacks

lenary:
These relate to issues found in https://github.com/llvm/llvm-project/pull/222850 

Aside: it seems the `tailcall` versions never landed in either LLVM or GCC upstream.


nemanjai (commented):
Perhaps it should be mentioned here that the `__riscv_save_<N>+` and `__riscv_save_cfiss_<N>+` routines should the defined in separate objects as the `_cfiss_` versions need to be compiled with the appropriate GNU note.

lenary:
Agreed.

The objects for the `cfiss` and `swss` versions need to be marked correctly.

src/toolchain-conventions.adoc
```diff
@@ -274,41 +274,59 @@ supported by GCC on X86 could be adopted.
...
+
+- `+__riscv_save_<N>+`
+- `+__riscv_save_cfiss_<N>+`
+- `+__riscv_save_swss_<N>+`
+- `+__riscv_restore_<N>+`
+- `+__riscv_restore_cfiss_<N>+`
+- `+__riscv_restore_swss_<N>+`
+- `+__riscv_restore_tailcall_<N>+` (LLVM/compiler-rt only)
+- `+__riscv_restore_tailcall_cfiss_<N>+` (LLVM/compiler-rt only)
+- `+__riscv_restore_tailcall_swss_<N>+` (LLVM/compiler-rt only)
```
kito-cheng: I am happy to adding those routines into libgcc as well (and targeting GCC 17) even GCC isn't support that code gen yet, so I would suggest drop `(LLVM/compiler-rt only)`

```suggestion
- `+__riscv_restore_tailcall_<N>+`
- `+__riscv_restore_tailcall_cfiss_<N>+`
- `+__riscv_restore_tailcall_swss_<N>+`
```

kito-cheng (commented):
I am little hesitate to provide those 3 variants since it should be handled by multilib in theory, e.g. libc must provide multilib for CFI or SW shadow stack, so that means all libgcc/compiler-rt should also provide multilib to make sure CFI and SW shadow stack can work correctly.

So...I am not sure provide all 3 variants for all multilib is reasonable to resolve this issue.

lenary:
While I understand that this "should be handled in theory", I'm not sure we have a way to indicate to the compiler that the save/restore routines are compatible with these codegen schemes.

It's quite common to use clang with an older glibc/gcc sysroot, and I don't want the shadow stack codegen assuming that the save/restore routines are doing something that they're not. Notably, if you use the wrong save/restore routines, then you get no protection in that function at all.

So with this in mind, changing the names means you definitely get a linker error if you use shadow stacks (either kind) with the save/restore routines from an old sysroot, as the new names are not available.

I suppose if we are marking the save/restore routine objects correctly, and using a new enough linker, then you would get the incompatible markings errors/warnings, if you had enabled them. LLD has these reports, but only for zicfiss and not for software shadow stack, and I don't know if binutils has them at all. So this doesn't seem very viable IMO.

Maybe we can document that a library may provide only one set of the save/restore routines, but the names still indicate what they actually do? I don't know if that creates other problems.

kito-cheng:
I thought about this again, and I think we need to discuss this problem in two stages: link-time checks for a single output ELF, and load-time checks for all dynamically loaded objects. These roughly match the statically linked runtime and dynamically linked runtime cases, but describing them by when the checks happen is more accurate.

Let me start with the easier link-time case. The static linker already has options such as `-z zicfi*` that can use GNU program properties to check CFI compatibility between input objects.

Software shadow stack currently has only a loose checking mechanism. We can use the `Tag_RISCV_x3_reg_usage` build attribute to check how x3 is used, but we do not have a complete mechanism like CFI to check compatibility and report errors. Maybe we should also define a GNU program property, or another ABI marker suitable for dynamic loading, for software shadow stack.

For a statically linked runtime, I believe this still needs to be handled through multilib. As long as the whole multilib, including other runtime functions in libgcc/compiler-rt, is built with the correct options, and we fix the `__riscv_save_<N>` / `__riscv_restore_<N>` implementations, there should be no additional problem.

Now back to the load-time part. When I saw your comment about using an older sysroot, I realized that I had not carefully considered the dynamically linked runtime case in my previous comment. Multilib can make sure we select the correct library at link time, but it cannot guarantee that the shared libraries loaded at runtime, including objects loaded later through `dlopen()`, have the same CFI/SW-SS properties.

My understanding is that one of the problems you want to solve is this: if a dynamically linked binary uses CFI or SW-SS, but finds a runtime without the same mechanism at link time or load time, how do we avoid silently using incompatible save/restore routines?

Adding the `_cfiss` / `_swss` suffixes is a very direct solution. If the runtime does not provide the required symbol, the static linker or dynamic loader will report an error instead of silently using the normal version.

However, I think this solves only a small part of the problem. If the program does not use the save/restore routines, or if the incompatible code is in other libgcc/compiler-rt functions, such as soft-float helpers, the problem will not be detected.

I can think of a few possible mechanisms, although they work at different levels:

1. Strong aliases

   We could continue with the naming scheme proposed in this PR. In a CFI-enabled runtime, `__riscv_save_cfiss_<N>` could be a strong alias of the corresponding `__riscv_save_<N>` implementation. This gives us two different symbol names without duplicating the implementation.

   However, this is mainly an implementation detail rather than a complete compatibility check. We would also need to confirm that the unsuffixed symbol and the suffixed implementation are backward-compatible. In particular, software shadow stack changes the use of x3, so we cannot simply assume that a normal program can safely use the same implementation.

2. Symbol versioning

   I personally do not like symbol versioning very much, but it is certainly a useful tool for checking ABI compatibility with dynamic linking.

   For example, code using CFI or SW-SS could reference:

   ```
   __riscv_save_<N>@RISCV_ZICFISS
   __riscv_save_<N>@RISCV_SWSS
   ```

   Normal code could use the default version (w/o version symbol):

   ```
   __riscv_save_<N>
   ```

   Existing programs could continue to use the default version, while programs using CFI or SW-SS would explicitly require the corresponding version. If the runtime does not provide that version, the linker or dynamic loader can report an error.

   However, like the suffix approach, symbol versioning can only check the ABI of the save/restore routines. It cannot prove that the other runtime functions in the same DSO have the same protection.

3. GNU program properties with dynamic-loader enforcement

   I think this is the mechanism that should check the compatibility of the whole shared object. The dynamic loader should check every DSO in the initial dependency closure, as well as objects loaded later through `dlopen()`.

   The glibc patch already provides a tunable to control the related behavior [2]. However, as far as I understand, the current implementation does not have a strict mode that requires every loaded object to have the CFI property and terminates the program otherwise. I believe adding such a strict/error policy should be possible.

I prefer option 3 as the main compatibility mechanism, since this is the kind of problem that GNU program properties are intended to solve. Options 1 and 2 can only be additional ABI checks for the save/restore routines. They cannot check the other functions in libgcc/compiler-rt.

If the dynamic loader can strictly check the GNU program property of every loaded object, the remaining question is whether it is still useful to add suffixes or symbol versions to the save/restore routines as a migration or defense-in-depth mechanism.

I am not sure if I have missed any other cases or limitations. :)

[1] https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/417
[2] https://patchwork.sourceware.org/project/glibc/patch/20260628070241.88310-10-jesse.huang@sifive.com/

NOTE: Written in Chines by me and translated by ChatGPT :P

